PR #20, opened 2022-05-30
feat: update AaveV2StablecoinCellar and CellarRouter

brianle83:
## Major Changes:
- Remove the inactive/active share model.
     - The problem it originally tried to solve (users depositing before their assets could be deployed to Aave and withdrawing to earn as much yield as someone who did) was (1) not as significant as we originally thought and could be easy mitigated, (2) had a solution that introduced more problems than it solved, and (3) was very gas inefficient.
- Cache value of `totalAssets()` (or at least a portion of it) which is updated with `accrue()`.
     - New yield is distributed linearly to the cellar across a period of time.
          - Protects against sandwich attacking `accrue()`.
     - Allows for more efficient method of accruing fees.
          - Old method was inefficient as a result of having to work around the rebasing nature of Aave aTokens. By caching the value, we no longer have this problem and as a result can implement a more optimal way of fee accrual.
- Refactor how fees were taken.
     - Previous implemention was inefficient, involved excessive external calls that were run everytime `activeAssets` was updated.
- Fix potentially criticial overflow error with `maxMint()` that would have arrisen if we lifted deposit and liquidity limits by setting them to `type(uint256).max`.
          
## Minor Changes:
- Update to Solidity version 0.8.13.
- Reduce platform fees from 1% annually to 0.25% annually.
- Add funcitonality to pull assets from the current lending position with `exitPosition()`.
- Inherit from ERC4626 instead of reimplementing it.
- Separate `setShutdown()` into `initiateShutdown()` and `liftShutdown()` to allow calls to start a shutdown vs lift a shutdown to be gated differently through Steward.
- Remove internal versions of accounting operations (eg. `_convertToShares()`, `_convertToAssets()`). 
- Rename `accrueFees()` to `accrue()` because it now does more than accrue fees.
- Rename `transferFees()` to `sendFees()`.
- Change `Deposit` and `Withdraw` events to conform to ERC4626.
- Remove `AccruedPerformanceFees` event and `AccruedPlatformFees` and combine them into `Accrued` event since they are always both accrued simultaneously now.
- Remove `getUserBalances()` and `getUserDeposits()` functions that are no longer necessary with the removal of the inactive/active share model.
- And lots of little gas optimizations.


contracts/AaveV2StablecoinCellar.sol
```diff
@@ -1,123 +1,215 @@
...
+        emit TrustChanged(address(position), trust);
+    }
+
+    // ======================================== LIMITS CONFIG ========================================
+
     /**
      * @notice Maximum amount of assets that can be managed by the cellar. Denominated in the same decimals
      *         as the current asset.
-     * @dev Set to `type(uint256).max` to have no limit.
+     * @dev Set to `type(uint128).max` to have no limit.
```
kkennis, line 145: A bit misleading no? There is still a limit, and since `2^128` is roughly equal to `10^38`, if a coin has 18 decimal places then we can only accept `1e20` units of coins. Worth noting that meme coins like shib have a total supply around `1e15`, so for future cellars, might be a little close for comfort (not an issue since we know all the stablecoins here)

contracts/AaveV2StablecoinCellar.sol
```diff
@@ -1,123 +1,215 @@
...
+     * @notice Prevent a function from being called during a shutdown.
+     */
+    modifier whenNotShutdown() {
+        if (isShutdown) revert STATE_ContractShutdown();
+
+        _;
+    }
+
+    /**
+     * @notice Shutdown the cellar. Used in an emergency or if the cellar has been depreciated.
```
kkennis, line 192: Typo:

```suggestion
     * @notice Shutdown the cellar. Used in an emergency or if the cellar has been deprecated.
```

contracts/AaveV2StablecoinCellar.sol
```diff
@@ -199,995 +291,560 @@ contract AaveV2StablecoinCellar is IAaveV2StablecoinCellar, ERC20, Ownable {
...
 
     /**
-     * @dev Same as `inactiveAssets` but forcibly denoted with 18 decimals of precision.
+     * @notice The total amount of locked yield still being distributed.
      */
-    function _inactiveAssets() internal view returns (uint256) {
-        uint256 assets = asset.balanceOf(address(this));
-        return assets.changeDecimals(assetDecimals, decimals);
-    }
+    function totalLocked() public view returns (uint256) {
```
kkennis, line 367: This is very elegant

contracts/AaveV2StablecoinCellar.sol
```diff
@@ -199,995 +291,560 @@ contract AaveV2StablecoinCellar is IAaveV2StablecoinCellar, ERC20, Ownable {
...
     /**
-    * @notice Simulate the effects of minting shares at the current block, given current on-chain
-    *         conditions.
+     * @notice Simulate the effects of minting shares at the current block, given current on-chain conditions.
      * @param shares amount of shares to mint
      * @return assets that will be deposited
      */
-    function previewMint(uint256 shares) public view returns (uint256) {
-        uint256 supply = totalSupply; // Saves an extra SLOAD if totalSupply is non-zero.
+    function previewMint(uint256 shares) public view override returns (uint256 assets) {
```
kkennis, line 415: Do we need the same logic for `previewDeposit`, or alternatively, do we need to override `deposit` from ERC4626 to disable it? I don't see how deposit is doing the same decimal accounting as `previewMint` does here

kkennis (commented):
Pretty beautiful, only kinda "code smell" thing would be that the `updateBalance` flags for the internal deposit and withdraw functions seem kind of error-prone/non-declarative, but I don't really have a better suggestion except maybe just always having to explicitly update the balance outside of deposit and withdraw

contracts/AaveV2StablecoinCellar.sol
brianle83: Yeah I thought this might be weird too, but was done to strictly adhere to ERC4626. They say about the `max` functions that they "MUST return 2 ** 256 - 1 if there is no limit on the maximum amount that may be deposited/minted".

I could fix this and make it less confusing by just not packing `liquidityLimit` and `depositLimit` into uint128's. Arguably it was a bit of an overzealous gas optimization.

contracts/AaveV2StablecoinCellar.sol
```diff
@@ -199,995 +291,560 @@ contract AaveV2StablecoinCellar is IAaveV2StablecoinCellar, ERC20, Ownable {
...
-    /**
-     * @notice Withdraws assets to receiver by redeeming shares from owner.
-     * @param assets amount of assets being withdrawn
-     * @param receiver address of account receiving the assets
-     * @param owner address of the owner of the shares being redeemed
-     * @return shares amount of shares redeemed
+     * @param receiver address that will receive minted shares
      */
-    function withdraw(
+    function beforeDeposit(
```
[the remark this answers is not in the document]
brianle83: It is called in `contracts/base/ERC4626.sol`, which it inherits from.

contracts/AaveV2StablecoinCellar.sol
```diff
@@ -178,7 +270,7 @@ contract AaveV2StablecoinCellar is IAaveV2StablecoinCellar, ERC20, Ownable {
      */
     constructor(
         ERC20 _asset,
-        address[] memory _approvedPositions,
+        ERC20[] memory _approvedPositions,
```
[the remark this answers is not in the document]
brianle83: When I've been deploying the cellar, I've been aware of this too and just didn't include `_asset` in _approvedPositions` because you are right it would be redundant.

brianle83:
> Pretty beautiful, only kinda "code smell" thing would be that the `updateBalance` flags for the internal deposit and withdraw functions seem kind of error-prone/non-declarative, but I don't really have a better suggestion except maybe just always having to explicitly update the balance outside of deposit and withdraw

Agree about `updateBalance`, I changed this [here](https://github.com/PeggyJV/cellar-contracts/pull/20/commits/c3ab5a205049f550c8c73c1775bfcd02c647b966).

sandbochs:
@brianle83 Can you elaborate a little bit on the inactive and active share model that was removed? I can't remember how we denoted the difference between the two.

The new arch still has the concept of 'active' vs 'inactive' liquidity right? Inactive being assets held by the contract, and active being deployed into the lendingPool.

I noticed totalAssets and totalHoldings are not converting to 18 decimal places anymore. Is that because everything is normalized before being saved?

brianle83:
> @brianle83 Can you elaborate a little bit on the inactive and active share model that was removed? I can't remember how we denoted the difference between the two.

Yeah, we would differentiate between inactive and active shares by comparing the `timeDeposited` timestamp that was stored with every deposit to `lastTimeEnteredPosition` to determine whether its shares were inactive or active. If `lastTimeEnteredPosition > timeDeposited` then the shares were considered active, otherwise, inactive.

To make this worked, we needed to store a `timeDeposited` with every deposit for each user. When a user withdrew, we'd retrieve and loop through all a user's deposits and determine how many active vs inactive shares they had. Inactive shares were redeemable for the exact amount originally deposited to mint those shares (ie. they earned no yield) while active shares were redeemable for the amount originally deposited + yield.
  
> The new arch still has the concept of 'active' vs 'inactive' liquidity right? Inactive being assets held by the contract, and active being deployed into the lendingPool.

Conceptually yes, but from the perspective of the cellar there is no longer any notion of inactive vs active shares. Each share is treated the same (ie. fungible). From a data perspective, you also don't have to worry about differentiating between inactive vs active shares like before. Now essentially all shares are "active".

> I noticed totalAssets and totalHoldings are not converting to 18 decimal places anymore. Is that because everything is normalized before being saved?

There was a need before to normalize all asset balances to 18 decimals because we'd store the assets deposited for each deposit (used to determine how much assets were withdrawn when attempting to redeem inactive shares) but this is no longer necessary without the inactive/active share model. Now the cellar stores, calculates, and returns all asset amounts using the same decimals of precision as the current `asset` (eg. if asset is USDC then 6 decimals, if FRAX then 18). This changes only when there is a rebalancing and the value of `asset` changes. When this occurs, previous asset accounting (eg. `totalBalance`, `totalAssets`, `maxLocked`) is scaled/truncated to the decimals of the new asset.

Whenever in doubt about the decimals of precision used for an asset amount, it will be the same as whatever `asset.decimals()` or `assetDecimals` was at that time (both are the same, `assetDecimals` is just a cached version of `asset.decimals()` stored for gas efficiency).  And shares, as before, will always be 18 decimals.

sandbochs:
@brianle83 thank you for the write up. It seems I will still need to convert decimals to 18 when snapshotting historical totalAssets to normalize the charts.

PR #39, opened 2022-06-20
feat(PriceRouter): add price router

### LINKED ISSUE #30: feat(PriceRouter): add PriceRouter module

crispymangoes:
Closes #30


src/PriceRouter.sol
```diff
@@ -0,0 +1,162 @@
...
+import { ERC20 } from "@solmate/tokens/ERC20.sol";
+import { SafeTransferLib } from "@solmate/utils/SafeTransferLib.sol";
+import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
+import { BaseAdaptor } from "./BaseAdaptor.sol";
+
+contract PriceRouter is Ownable {
+    using SafeTransferLib for ERC20;
+
+    //in terms of 8 decimal USD
+    //determined by interfacing with adaptors
```
0xble, line 13: What does this mean? Could you add more detail to this comment?

src/ChainlinkPriceFeedAdaptor.sol
```diff
@@ -0,0 +1,64 @@
...
+    constructor(FeedRegistryInterface _feedRegistry) {
+        feedRegistry = _feedRegistry;
+    }
+
+    //TODO should this revert if it can't find USD or ETH price?
+    function getPricingInformation(address baseAsset)
+        external
+        view
+        returns (PriceRouter.PricingInformation memory info)
+    {
```
0xble, line 37: Try to reduce the amount of external calls in this function, eg. fetch max and min values with one call if possible, fetch latest answer and latest timestamp with one call.

crispymangoes:
It doesn't look like solidity will revert when doing explicit type conversions, even if the conversion would lose data, or change the sign. So conversions should be handled by internal functions that check to make sure we aren't losing data or changing sign.

PR #42, opened 2022-06-23
audit(AaveV2StablecoinCellar): anchor performance fee accounting

0xble:


0xble:
> How is the highwater mark affected by the withdrawal of liquidity by cellar users?

It does not count entering and exiting positions as gains/losses (different from user deposits and withdraws to/from the holding pool, which should have no effect because we only take performance fees from the lending position on Aave).

When entering/exiting, we add/subtract from the watermark.

PR #52, opened 2022-07-05
Feat/(SwapRouter) add curve and balancer

zpoken:


crispymangoes:
- Remove lines 15,16 in SwapRouter.sol as this PR adds those exchanges
- modify `swapWithCurve` to accept a `recipient` address like how `swapWithUniV2` does it. This done so that we can better gate keep where swapped funds are being sent to.
- modify `swapWithBalancerV2` adding the `recipient` address like above.

Other than those small changes, it looks pretty good to me!

crispymangoes:
Thanks for adding in the recipient address! Is it possible for the recipient address to be passed with the rest of the swap information, so that the swap sends the funds directly to the recipient? Or is the transfer at the end of the functions required? 

Also go ahead and remove the `from` variables from the swap data and use `msg.sender` instead. If the caller is allowed to specify the `from` address, then any user/contract that has approved the swap router to spend their funds is at risk of losing them.

zpoken:
> Thanks for adding in the recipient address! Is it possible for the recipient address to be passed with the rest of the swap information, so that the swap sends the funds directly to the recipient? Or is the transfer at the end of the functions required?

Unfortunately, function balancerExchangeProxy.multihopBatchSwapExactIn sends funds exclusively to msg.sender.
In function curveRegistryExchange.exchange_multiple we can add _receiver to the interface. Ok.

zpoken:
> Also go ahead and remove the from variables from the swap data and use msg.sender instead. If the caller is allowed to specify the from address, then any user/contract that has approved the swap router to spend their funds is at risk of losing them.

Ok

PR #55, opened 2022-07-07
perf(Cellar): struct pack PositionData by converting highWatermark fr…

crispymangoes:
…om int256 to int248

### DIFF
--- a/src/base/Cellar.sol
+++ b/src/base/Cellar.sol
@@ -75,13 +75,14 @@ contract Cellar is ERC4626, Ownable, Multicall {
 
     /**
      * @notice Data related to a position.
+     * @dev struct pack a uint8(positionType) with a int248(highWatermark)
      * @param positionType value specifying the interface a position uses
      * @param highWatermark amount representing the balance this position needs to exceed during the
      *                      next accrual to receive performance fees
      */
     struct PositionData {
         PositionType positionType;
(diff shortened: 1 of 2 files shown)


src/base/Cellar.sol
```diff
@@ -75,13 +75,14 @@ contract Cellar is ERC4626, Ownable, Multicall {
 
     /**
      * @notice Data related to a position.
+     * @dev struct pack a uint8(positionType) with a int248(highWatermark)
```
0xble, line 78: Can delete this comment.

src/base/Cellar.sol
```diff
@@ -938,7 +939,7 @@ contract Cellar is ERC4626, Ownable, Multicall {
             uint256 balanceThisAccrual = positionBalances[i];
 
             // Measure yield earned against this position's high watermark.
-            int256 yield = balanceThisAccrual.toInt256() - positionData.highWatermark;
+            int256 yield = balanceThisAccrual.toInt256() - positionData.highWatermark; // highWatermark is implicitly converted to an int256
```
0xble, line 942: Can delete this comment too. For future code comments though make sure to capitalize and use periods.

src/base/Cellar.sol
crispymangoes: Done1

PR #58, opened 2022-07-21
Feat/add share price high watermark

crispymangoes:
Still need to test the platform fee logic in `sendFees`, but the new high watermark logic has been tested and works as expected. 

Could you look at the preview functions? In order to account for the new performance fee logic, the code reduces the amount of assets passed into the preview functions in order to represent the devaluation of share price when performance fees are minted.

Ideally I wanted to change the total supply value in the preview functions, but the functions were set up to accept the total assets, so doing it this way changed the least amount of code.


crispymangoes:
Added sendFees, test, and discussed preview function with Brian

0xble:
Reduce the state variables this cellar has down to a maximum of 15. It is currently at 17.

Considering there are a lot of state variables being used to track fees (eg. `platformFee`, `performanceFee`, `strategistPerformanceCut`, `strategistPayoutAddress`, `sharePriceHighWatermark`, etc.), consider moving these into a struct (eg. `FeeData`) to store it all in a single state variable.

src/base/Cellar.sol
```diff
@@ -76,12 +78,9 @@ contract Cellar is ERC4626, Ownable, Multicall {
     /**
      * @notice Data related to a position.
      * @param positionType value specifying the interface a position uses
-     * @param highWatermark amount representing the balance this position needs to exceed during the
-     *                      next accrual to receive performance fees
      */
     struct PositionData {
```
0xble, line 82: Since this struct now only stores a single value, delete this struct and change `getPositionData` to `mapping(address => PositionType) getPositionType`.

src/base/Cellar.sol
```diff
@@ -1069,6 +1051,103 @@ contract Cellar is ERC4626, Ownable, Multicall {
...
+    /**
+     * @notice Stores the share price to be used as a High Watermark to calculate performance fees
+     */
+    uint256 public sharePriceHighWatermark;
+
+    /**
+     * @notice Number 0 -> 1e18 defining how much of the performance fee goes to the strategist
+     */
+    uint256 public strategistPerformanceCut = 0.75e18;
+
```
0xble, line 1063: Organize these state variables under the "FEE CONFIG" section.

src/base/Cellar.sol
```diff
@@ -1069,6 +1051,103 @@ contract Cellar is ERC4626, Ownable, Multicall {
...
+            return 0;
+        }
+    }
+
+    /**
+     * @notice Mints cellar performance fee shares if current share price is above high watermark
+     * @param _totalAssets uint256 value of the total assets in the cellar
+     */
+    function _takePerformanceFees(uint256 _totalAssets) internal {
+        if (performanceFee == 0) return;
```
0xble, line 1119: In order to 1) reduce the possible sources of error causing discrepencies between the amounts returned by deposit/withdraw functions and their `preview` counterparts and 2) reduce the amount of redundant code between `calculatePerformanceFee` and `_takePerformanceFees`, `_takePerformanceFees` and `_calculatePerformanceFees` should basically be the same except that `_takePerformanceFees` actually mints the fees while `_calculatePerformanceFees` just returns how much will be minted without actually minting them.

Also rename `_calculatePerformanceFees` to `_previewPerformanceFees`.

src/base/Cellar.sol
```diff
@@ -1079,23 +1158,67 @@ contract Cellar is ERC4626, Ownable, Multicall {
...
+        uint256 _totalAssets = totalAssets();
+
+        // Since this action mints shares, calculate outstanding performance fees due.
+        _takePerformanceFees(_totalAssets);
+
+        uint256 currentBalance = balanceOf[address(this)];
+
+        console.log("Performance Fee Assets", _convertToAssets(currentBalance, _totalAssets)); // This is right
+
+        uint256 strategistFeeSharesDue = currentBalance.mulDivDown(strategistPerformanceCut, 1e18);
```
0xble, line 1174: Use `mulWadDown` instead of `mulDivDown` for all the instances where you are dividing by `1e18`.

I counted three places (including this one) where this should be changed.

src/base/Cellar.sol
```diff
@@ -13,6 +13,8 @@ import { IGravity } from "../interfaces/IGravity.sol";
 import { AddressArray } from "src/utils/AddressArray.sol";
 import { Math } from "../utils/Math.sol";
 
+//TODO remove this
+import { Test, console } from "@forge-std/Test.sol";
```
0xble, line 17: After you are done with testing, don't forget to clean up `console.log`s and delete this import.

src/base/Cellar.sol
```diff
@@ -1069,6 +1051,104 @@ contract Cellar is ERC4626, Ownable, Multicall {
...
+     * @notice Number 0 -> 1e18 defining how much of the platform fee goes to the strategist
+     */
+    uint256 public strategistPlatformCut = 0.75e18;
+
+    /**
+     * @notice Address to send the strategists fee shares
+     */
+    address public strategistPayoutAddress;
+
+    ///@dev resets high watermark to current share price
```
0xble, line 1074: Delete `resetHighWatermark` for now.

Also when we add it back, use `convertToAssets` to get the current share price instead of how its implemented here.

src/base/Cellar.sol
```diff
@@ -372,41 +361,90 @@ contract Cellar is ERC4626, Ownable, Multicall {
...
+            strategistPayoutAddress: address(0),
+            platformFee: 0.01e18,
+            performanceFee: 0.1e18,
+            feesDistributor: hex"000000000000000000000000b813554b423266bbd4c16c32fa383394868c1f55"
+        });
+
+    /**
+     * @notice Get the fee data values for this cellar
+     */
+    function getFeeData() external view returns (FeeData memory) {
```
0xble, line 426: Delete this.

A getter function is automatically created for `feeData` by making it `public`.

src/base/Cellar.sol
```diff
@@ -416,9 +454,32 @@ contract Cellar is ERC4626, Ownable, Multicall {
...
     function setFeesDistributor(bytes32 newFeesDistributor) external onlyOwner {
-        emit FeesDistributorChanged(feesDistributor, newFeesDistributor);
+        emit FeesDistributorChanged(feeData.feesDistributor, newFeesDistributor);
+
+        feeData.feesDistributor = newFeesDistributor;
+    }
+
+    ///@dev Sets the Strategists cut of performance fees
+    function setStrategistPerformanceCut(uint64 cut) external onlyOwner {
+        require(cut <= 1e18, "Invalid cut");
```
0xble, line 464: Change all `require` statements to custom errors.

src/base/Cellar.sol
```diff
@@ -416,9 +454,32 @@ contract Cellar is ERC4626, Ownable, Multicall {
      * @param newFeesDistributor formatted address of the new fee distributor module
      */
     function setFeesDistributor(bytes32 newFeesDistributor) external onlyOwner {
-        emit FeesDistributorChanged(feesDistributor, newFeesDistributor);
+        emit FeesDistributorChanged(feeData.feesDistributor, newFeesDistributor);
+
+        feeData.feesDistributor = newFeesDistributor;
+    }
+
+    ///@dev Sets the Strategists cut of performance fees
```
0xble, line 462: Change this to the `@notice` tag instead of `@dev`.

src/base/Cellar.sol
```diff
@@ -372,41 +361,90 @@ contract Cellar is ERC4626, Ownable, Multicall {
...
 
     /**
-     * @notice Cosmos address of module that distributes fees, specified as a hex value.
-     * @dev The Gravity contract expects a 32-byte value formatted in a specific way.
+     * @notice Stores all fee data for cellar.
      */
-    bytes32 public feesDistributor = hex"000000000000000000000000b813554b423266bbd4c16c32fa383394868c1f55";
+    FeeData public feeData =
+        FeeData({
+            sharePriceHighWatermark: 0,
```
0xble, line 414: You can just rename this to `highWatermark`.

src/base/Cellar.sol
```diff
@@ -572,6 +633,9 @@ contract Cellar is ERC4626, Ownable, Multicall {
         // `accrue` will take platform fees from 1970 to the time it is called.
         lastAccrual = uint64(block.timestamp);
 
+        // Initialize SharePriceHighWatermark
+        feeData.sharePriceHighWatermark = 10**asset.decimals();
```
0xble, line 637: Refactor to `10**_asset.decimals()` instead, it's more gas efficient because it uses an `MLOAD` instead of an `SLOAD`.

src/base/Cellar.sol
```diff
@@ -598,6 +662,44 @@ contract Cellar is ERC4626, Ownable, Multicall {
         _depositTo(holdingPosition, assets);
     }
 
+    function deposit(uint256 assets, address receiver) public override returns (uint256 shares) {
+        uint256 _totalAssets = totalAssets();
+
+        _takePerformanceFees(_totalAssets);
+        // Check for rounding error since we round down in previewDeposit.
+        require((shares = _convertToShares(assets, _totalAssets)) != 0, "ZERO_SHARES");
```
0xble, line 670: I know this is just from ERC4626, but could you use refactor this into a custom error using `USR_ZeroShares`?

src/base/Cellar.sol
```diff
@@ -1069,6 +1116,58 @@ contract Cellar is ERC4626, Ownable, Multicall {
...
+        internal
+        view
+        returns (uint256 feeInAssets, uint256 currentSharePrice)
+    {
+        if (feeData.performanceFee == 0 || _totalAssets == 0) return (0, 0);
+        currentSharePrice = _convertToAssets(10**decimals, _totalAssets);
+        if (feeData.sharePriceHighWatermark < currentSharePrice) {
+            //find how many assets make up the fee
+            uint256 yield = ((currentSharePrice - feeData.sharePriceHighWatermark) * totalSupply) / 10**decimals;
+            feeInAssets = yield.mulDivDown(feeData.performanceFee, 1e18);
```
0xble, line 1135: Optimize this by storing `feeData.performanceFee` and `feeData.sharePriceHighWatermark` and reading from that instead of reading from state redundantly.

src/base/Cellar.sol
```diff
@@ -1069,6 +1116,58 @@ contract Cellar is ERC4626, Ownable, Multicall {
...
+    function _previewPerformanceFees(uint256 _totalAssets)
+        internal
+        view
+        returns (uint256 feeInAssets, uint256 currentSharePrice)
+    {
+        if (feeData.performanceFee == 0 || _totalAssets == 0) return (0, 0);
+        currentSharePrice = _convertToAssets(10**decimals, _totalAssets);
+        if (feeData.sharePriceHighWatermark < currentSharePrice) {
+            //find how many assets make up the fee
+            uint256 yield = ((currentSharePrice - feeData.sharePriceHighWatermark) * totalSupply) / 10**decimals;
```
0xble, line 1134: Use `mulDivDown` here.

src/base/Cellar.sol
```diff
@@ -1079,23 +1178,53 @@ contract Cellar is ERC4626, Ownable, Multicall {
...
+     * @dev assumes cellar's accounting asset is able to be transferred and sent to Cosmos
      */
     function sendFees() public onlyOwner {
+        require(feeData.strategistPayoutAddress != address(0), "Strategist Payout not set");
+        uint256 _totalAssets = totalAssets();
+
+        // Since this action mints shares, calculate outstanding performance fees due.
+        _takePerformanceFees(_totalAssets);
+
+        uint256 currentBalance = balanceOf[address(this)];
```
0xble, line 1190: Rename this to `totalFees`.

src/base/Cellar.sol
```diff
@@ -1079,23 +1178,53 @@ contract Cellar is ERC4626, Ownable, Multicall {
...
+        }
+
+        emit Transfer(address(this), feeData.strategistPayoutAddress, strategistFeeSharesDue);
+
+        lastAccrual = uint32(block.timestamp);
+
         // Redeem our fee shares for assets to send to the fee distributor module.
-        uint256 totalFees = balanceOf[address(this)];
-        uint256 assets = previewRedeem(totalFees);
+        currentBalance = balanceOf[address(this)];
```
0xble, line 1216: Optimize this to not read `balanceOf[address(this)]` redundantly.

src/base/Cellar.sol
```diff
@@ -1079,23 +1178,53 @@ contract Cellar is ERC4626, Ownable, Multicall {
     /**
      * @notice Transfer accrued fees to the Sommelier chain to distribute.
      * @dev Fees are accrued as shares and redeemed upon transfer.
+     * @dev assumes cellar's accounting asset is able to be transferred and sent to Cosmos
      */
     function sendFees() public onlyOwner {
```
0xble, line 1183: Optimize this by caching some storage variables instead of reading them twice.

src/base/Cellar.sol
```diff
@@ -598,6 +690,59 @@ contract Cellar is ERC4626, Ownable, Multicall {
...
+
+    function deposit(uint256 assets, address receiver) public override returns (uint256 shares) {
+        uint256 _totalAssets = totalAssets();
+
+        _takePerformanceFees(_totalAssets);
+
+        // Check for rounding error since we round down in previewDeposit.
+        if ((shares = _convertToShares(assets, _totalAssets)) == 0) revert USR_ZeroShares();
+
+        beforeDeposit(assets, shares, receiver);
```
kkennis, line 715: Seems like you could keep things DRY and replace lines 715-722 with `mint(shares, receiver)`?

kkennis: Same opportunity for `redeem`/`withdraw` (maybe a little more work there, you might have to create a shared internal function to prevent repeated calls to `_getData`)

src/base/Cellar.sol
```diff
@@ -1069,6 +1165,67 @@ contract Cellar is ERC4626, Ownable, Multicall {
...
+     */
+    event HighWatermarkReset();
+
+    /**
+     * @notice Resets High Watermark to equal current total assets.
+     * @notice This function can be abused by Strategists, so it should only be callable by governance.
+     */
+    function resetHighWatermark() external onlyOwner {
+        feeData.highWatermark = totalAssets();
+        emit HighWatermarkReset();
```
kkennis, line 1179: Add a param for the new `totalAssets` value?

kkennis:
Code looks great and is well-implemented as far as intention goes! Would still like to be able to run `forge coverage` to see if the tests are missing anything, but doesn't work :|

Only general comment I have is that I'm not sure that the high-watermark model is the optimal solution for every flavor of cellar forever, especially one that takes more directional positions. By the very nature of the name we know that the high-watermark model takes fees on gains (passing the prev high watermark) and does nothing on losses. This is OK for a cellar that doesn't expect a loss (like the AAVE cellar). But sooner rather than later, there will probably need to be some cellars whose fee model is such that the fees are _lost_ on losses. 

Let's say the performance fee is 10%, and the cellar buys $1mm of $XYZ at $1, rides it to $3, and then $XYZ falls to $0.50 without the cellar closing that position. What happens is:

* Performance fee accrued to owners is $200,000 (10% of $2mm profit)
* Users whose deposits contributed to that $1mm are now down **70%**!! (basically, their 1mm tokens of XYZ are now worth $500,000, _and_ $200,000 of that has been reserved for fees - leaving them with $300,000).

You can see how this is not a good solution and can reward managers for losing money, taking undue risk (why not buy a pump-and-dump if you keep your share of the "paper" profits and don't need to worry about selling the top). It's basically always setting up the fee taker to "sell the top" at the expense of the depositors - if the position is closed below the "top" - the high watermark, the **entire loss** falls on the users, and the fee takers made out as if they sold the exact top. I think we're going to need to re-think this before too long, as a user I'd not deposit into a "managed cellar" with this fee structure.

src/base/Cellar.sol
crispymangoes: Though this is set by governance, we'd want this to be accurate. If the reset value is above the current assets, the strategist has to make up that ground before earning fees. If the reset value is below, then the strategist earns performance fees for free.

kkennis: I mean for the event

src/base/Cellar.sol
```diff
@@ -288,8 +299,14 @@ contract Cellar is ERC4626, Ownable, Multicall {
...
         isTrusted[position] = false;
 
-        // Remove position from the list of positions if it is present.
-        positions.remove(position);
+        // Only remove position if it is not being used, is empty, and if it is not the holding position.
+        if (isPositionUsed[position]) {
+            uint256 positionBalance = _balanceOf(position);
+            if (positionBalance > 0) revert Cellar__PositionNotEmpty(position, positionBalance);
+            if (position == holdingPosition) revert Cellar__RemoveHoldingPosition();
+            positions.remove(position);
```
0xble, line 307: This will cause distrust to revert if the position being distrusted has a balance. I don't think this is what we want. It should still be distrusted, but just not removed from the `positions` array (otherwise it's balance will not be accounted for in `totalAssets`).

PR #60, opened 2022-08-08
test(SwapRouter): add full testing to `SwapRouter`

0xble:


crispymangoes:
Could you change these two tests to use vm.expectRevert instead of test fail?
`testFailSwapWithoutEnoughAssets`
`testFailSwapWithoutEnoughApproved`

crispymangoes:
Add test where an invalid Exchange is given to the swap function.

Should we add a revert error message if receiver is the zero address? I am not sure if all tokens revert on transfers to zero address. They should but people have done weird stuff with ERC20s before where transfers won't revert they just return false.

crispymangoes:
Add a vm.expectRevert test to get the Swap reverted message. I think if the path variable had an invalid token address that would happen. Also change this `revert("Swap reverted");` into an error.

0xble:
> Add test where an invalid Exchange is given to the swap function.
> 
> Should we add a revert error message if receiver is the zero address? I am not sure if all tokens revert on transfers to zero address. They should but people have done weird stuff with ERC20s before where transfers won't revert they just return false.

It's not possible to test, Solidity won't let you do this. It's typed to the `Exchange` enum. Although I get what you're larger concern is.

Turns out it's start enough to revert if you try to pass in an enum index/value that is out of range. I tested this using Solidity (won't even compile and show an error if you attempt)and using JavaScript through Hardhat (will revert).

0xble:
> Add test where an invalid Exchange is given to the swap function.
> 
> Should we add a revert error message if receiver is the zero address? I am not sure if all tokens revert on transfers to zero address. They should but people have done weird stuff with ERC20s before where transfers won't revert they just return false.

Regarding adding a check for zero address, we could. Though I tend to agree with t11s philosophy of not increasing the gas cost for everyone else by adding a check to prevent a few people from burning their tokens or being dumb. A motivated person could also just send their tokens to literally any other dead address (eg. `address(1)`) and achieve the same effect.

PR #78, opened 2022-11-01
Add fix to swap router, and add integration test

crispymangoes:


script/DeploySwapRouterWithApprovalFix.s.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.16;
+
+import "forge-std/Script.sol";
+import { SwapRouter } from "src/modules/swap-router/SwapRouter.sol";
+import { IUniswapV2Router02 as IUniswapV2Router } from "src/interfaces/external/IUniswapV2Router02.sol";
+import { IUniswapV3Router } from "src/interfaces/external/IUniswapV3Router.sol";
+
+contract DeploySwapRouterWithApprovalFixScript is Script {
+    address private uniswapV2Router = 0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D;
```
kkennis, line 10: Would there be a way to dynamically set these? Just thinking about all the multi-network deploy stuff we talked about yesterday....not sure the foundry best practice for deploying with different addresses based on different networks

PR #84, opened 2022-11-22
Feat/cellar optimization

crispymangoes:


src/base/Cellar.sol
```diff
@@ -1408,12 +1443,12 @@ contract Cellar is ERC4626, Owned, ERC721Holder {
     }
 
     /**
-     * @notice Get all the positions underlying assets.
+     * @notice Get all the creit positions underlying assets.
```
kkennis, line 1446: Typo - should be `credit`

kkennis (approved):

PR #101, opened 2023-02-10
Fix/euler q1 q2

crispymangoes:


src/modules/adaptors/Euler/EulerBaseAdaptor.sol
```diff
@@ -0,0 +1,71 @@
...
+ * @notice Implements shared logic/constants between EulerETokenAdaptor and EulerDebtTokenAdaptor.
+ * @author crispymangoes
+ */
+contract EulerBaseAdaptor {
+    using Math for uint256;
+
+    /**
+     * @notice Attempted to use an invalid subAccountId.
+     */
+    error EulerETokenAdaptor__InvalidSubAccountId();
```
exp7l, line 17: Consider using "EulerBaseAdaptor__InvalidSubAccountId" instead.

PR #111, opened 2023-04-19
WIP: Feat/price router v2

crispymangoes:


0xEinCodes (changes requested):

0xEinCodes (changes requested):
Nice work, I think this will be a cool upgrade that will open more doors for pricing! Overall the design makes sense to me:

- Extensibility for different, and even more exotic, price sources for future assets coming into the Sommelier protocol
- Keeping two main price sources available within the actual `PriceRouter.sol` contract allows for simpler price querying for the most common, for now, methodologies (Chainlink && UniswapV3 TWAP).

I made some comments throughout to better understand the design in finer detail, and some nit-picky reformatting stuff. We can discuss all this as well synchronously. Perhaps adding a list of TODOs at the beginning of this PR description would help line up the intent of the PR. From my understanding, you'll need to write tests still too. I've gotta get my head wrapped around some of the PriceRouter workflows so I may add some more comments in the coming days or so still.

Other Notes:
- I didn't check the Balancer docs in full or other external protocol docs to ensure that the calculations were carried out in the proper manner wrt to the respective protocol. I can go back in detail to do so though once tests are written, etc.
- I'd still like to look through, but just taking a break for now since there's a lot to this PR:
   - `WstEthExtension.t.sol,` `PriceRouter.t.sol,` `WstEthExtension.sol,` `CurveV1Extension.sol,` and `CurveV2Extension.sol,` and `UniswapV3Pool.sol` (the last three I presume are more-so from `PriceRouterv1.sol`

src/modules/price-router/PriceRouter.sol
```diff
@@ -160,30 +168,37 @@ contract PriceRouter is Ownable, AutomationCompatibleInterface {
...
+        // (uint8 derivative, bytes memory derivativeCheckData) = abi.decode(checkData, (uint8, bytes));
+        // if (derivative == 2) {
+        //     (upkeepNeeded, performData) = _checkVirtualPriceBound(derivativeCheckData);
+        // } else if (derivative == 3) {
+        //     (upkeepNeeded, performData) = _checkVirtualPriceBound(derivativeCheckData);
+        // } else revert PriceRouter__UnkownDerivative(derivative);
     }
 
     /**
      * @notice `performUpkeep` is set up to allow for multiple derivatives to use Chainlink Automation.
```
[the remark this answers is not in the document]
crispymangoes: I think we will do without it. There are a couple unique cases where having an automation job running helps with security, but we can implement that in extensions.

src/modules/price-router/PriceRouter.sol
```diff
@@ -3,28 +3,48 @@ pragma solidity 0.8.16;
...
 
 /**
  * @title Sommelier Price Router
  * @notice Provides a universal interface allowing Sommelier contracts to retrieve secure pricing
  *         data from Chainlink.
- * @author crispymangoes, Brian Le
+ * @author crispymangoes
  */
-contract PriceRouter is Ownable, AutomationCompatibleInterface {
+contract PriceRouter is Ownable {
```
0xEinCodes, line 22: A general question for the `PriceRouter` design:

What is the current design's response to oracles failing? Let's say the stETH : USD oracle goes down. Obviously, a lot of bad things will happen to the rest of DeFi, but it would be ideal for the Sommelier setup to isolate and mitigate contagion. 

Do we currently just have everything revert related to rebalances? I guess the assets themselves are still in the cellar possession, and rebalances just won't work. I'm basically wondering what our stance is on fallback and contingent oracle sources since there is some risk vector when working with just one oracle source (Chainlink) for example for an asset price. There's pro's and con's to fallback design vs not, and so I maybe we should open this as an issue to discuss.

wydt?

src/modules/price-router/PriceRouter.sol
```diff
@@ -41,7 +61,109 @@ contract PriceRouter is Ownable, AutomationCompatibleInterface {
...
+     */
+    address public pendingOwner;
+
+    /**
+     * @notice The starting time stamp of the transition.
+     */
+    uint256 public transitionStart;
+
+    /**
+     * @notice Allows Zero Id address to set a new owner, after the transition period is up.
```
0xEinCodes, line 122: Who is zero id?

src/modules/price-router/PriceRouter.sol
```diff
@@ -41,7 +61,109 @@ contract PriceRouter is Ownable, AutomationCompatibleInterface {
...
+
+    /**
+     * @notice Attempted to call `completeTransition` from an address that is not the pending owner.
+     */
+    error PriceRouter__OnlyCallableByPendingOwner();
+
+    /**
+     * @notice The amount of time it takes for an ownership transition to work.
+     */
+    uint256 public constant TRANSITION_PERIOD = 7 days;
```
0xEinCodes, line 109: Why 7 days?

test/Registry.t.sol
```diff
@@ -22,7 +22,7 @@ contract RegistryTest is Test {
     address private USDC_USD_FEED = 0x8fFfFfd4AfB6115b954Bd326cbe7B4BA576818f6;
 
     function setUp() external {
-        priceRouter = new PriceRouter();
+        priceRouter = new PriceRouter(registry);
         registry = new Registry(gravityBridge, swapRouter, address(priceRouter));
```
0xEinCodes, line 26: Another design concept to confirm:
This `PriceRouterV2` will be deployed and ideally replace the current `SLOT` for the `PriceRouterV1`, thereby updating Cellars querying for price, right? 

How does that transition look? Is it something along the lines of: 

-  PriceRouterV2 approved via audit, finalization, and governance.
- `Registry` simply updates the `PRICE_ROUTER_REGISTRY_SLOT` with new `PriceRouterV2`?

Is there an iterative approach to updating things?

test/testPriceRouter/PriceRouter.t.sol
```diff
@@ -0,0 +1,788 @@
...
+                    address(USDC),
+                    timestamp + 1 days,
+                    1 days
+                )
+            )
+        );
+        priceRouter.getValue(USDC, 1e6, USDC);
+    }
+
+    function testETHtoUSDPriceFeedIsChecked() external {
```
0xEinCodes, line 328: Do you think adding another test checking this same thing against different types of `derivatives` for `baseAsset` vs `quoteAsset`?

test/testPriceRouter/PriceRouter.t.sol
```diff
@@ -0,0 +1,788 @@
...
+        PriceRouter.AssetSettings memory settings = PriceRouter.AssetSettings(CHAINLINK_DERIVATIVE, BOND_ETH_FEED);
+        PriceRouter.ChainlinkDerivativeStorage memory stor = PriceRouter.ChainlinkDerivativeStorage(
+            100e18,
+            0.0001e18,
+            2 days,
+            true
+        );
+
+        // Simulate calling startEditAsset.
+        bytes32 editHash = keccak256(abi.encode(BOND, settings, abi.encode(stor)));
```
0xEinCodes, line 340: Why are we simulating `startEditAsset()` throughout the tests vs just using it? Arguably, there should be tests that implement this as that is the true happy path vs using storage manipulation with foundry right?

test/testPriceRouter/PriceRouter.t.sol
```diff
@@ -0,0 +1,788 @@
...
+        // If an asset is pending edit, it can be cancelled.
+        priceRouter.startEditAsset(USDC, settings, abi.encode(0));
+
+        priceRouter.cancelEditAsset(USDC, settings, abi.encode(0));
+    }
+
+    // ======================================= PRICING TESTS =======================================
+
+    function testExchangeRate() external {
+        PriceRouter.AssetSettings memory settings = PriceRouter.AssetSettings(CHAINLINK_DERIVATIVE, BOND_ETH_FEED);
```
0xEinCodes, line 571: Is it worthwhile to test the various different derivative types, or is that captured in Extension tests more-so?

test/testPriceRouter/PriceRouter.t.sol
```diff
@@ -0,0 +1,788 @@
...
+                    address(WETH),
+                    timestamp + 3600,
+                    3600
+                )
+            )
+        );
+        priceRouter.getValue(BOND, 1e18, USDC);
+    }
+
+    function testAddingATwapAsset() external {
```
0xEinCodes, line 382: Repeating my inquiry at the bottom of this test contract, but I think it would be worthwhile to test various derivative's exchange pricing.

0xEinCodes: That would add more testing to the Twap asset setup and pricing

src/interfaces/external/IBalancerPool.sol
```diff
@@ -0,0 +1,62 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity >=0.8.0;
+
+interface IBalancerPool {
```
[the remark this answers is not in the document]
crispymangoes: They are there if we need them later, but we don't explicitly need them right now.

src/interfaces/external/IBalancerPool.sol
```diff
@@ -0,0 +1,62 @@
...
+
+    function swapExactAmountOut(
+        address tokenIn,
+        uint256 maxAmountIn,
+        address tokenOut,
+        uint256 tokenAmountOut,
+        uint256 maxPrice
+    ) external returns (uint256 tokenAmountIn, uint256 spotPriceAfter);
+
+    function joinswapExternAmountIn(
```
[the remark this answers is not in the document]
crispymangoes: These function names are what the actual mainnet contracts call them, so we can not change their name, otherwise our calls would revert.

I dont think we will use this interface function, but if we ended up using it, and we changed the function name to camel case, then our calls would revert because the real mainnet contract function is called `joinswapExternalAmountIn ` not `joinSwapExternalAmountIn `.

src/modules/price-router/PriceRouter.sol
```diff
@@ -8,9 +8,13 @@ import { IChainlinkAggregator } from "src/interfaces/external/IChainlinkAggregat
...
 import { Math } from "src/utils/Math.sol";
 import { Address } from "@openzeppelin/contracts/utils/Address.sol";
-import { ICurvePool } from "src/interfaces/external/ICurvePool.sol";
-import { IAaveToken } from "src/interfaces/external/IAaveToken.sol";
+import { Extension } from "src/modules/price-router/Extensions/Extension.sol";
 
+import { UniswapV3Pool } from "src/interfaces/external/UniswapV3Pool.sol";
+import { OracleLibrary } from "@uniswapV3P/libraries/OracleLibrary.sol";
+
+// TODO so we could have some external watch dog contract that can freeze assets if there are issues detected
```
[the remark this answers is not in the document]
crispymangoes: Not at the moment, we could in the future but doing so will increase gas costs further. 

It is very important for lending protocols to have backup oracles so that liquidations can always happen, but in our case it isn't nearly as important. Yes if some assets chainlink feed were reverting, the multisig could brick cellars trying to price that asset by doing nothing, but if that happens, governance has the ability to force-ably remove the old multisig and replace it with one that will actually fix the problem.

src/modules/price-router/PriceRouter.sol
```diff
@@ -602,476 +616,48 @@ contract PriceRouter is Ownable, AutomationCompatibleInterface {
...
-        uint32 _posDelta,
-        uint32 _negDelta,
-        uint32 _rateLimit
-    ) external onlyOwner {
-        VirtualPriceBound storage vpBound = getVirtualPriceBound[_asset];
-        vpBound.posDelta = _posDelta;
-        vpBound.negDelta = _negDelta;
-        vpBound.rateLimit = _rateLimit == 0 ? DEFAULT_RATE_LIMIT : _rateLimit;
-    }
+    function _setupPriceForTwapDerivative(ERC20 _asset, address _source, bytes memory _storage) internal {
```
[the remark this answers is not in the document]
crispymangoes: No this contract will only support UniV3 TWAPS. In the future we might add curve support, if that opens up a lot more opportunities for us, but in the mean time I don't think it is needed.

src/modules/price-router/PriceRouter.sol
crispymangoes: Yes we currently have everything revert. Which is the safest thing for us. It is kinda rough for lending protocols that want to use our Cellar shares, but that is why setting up balancer 80/20 pools would be nice for them, since they can emergency liquidate people using the pool.

src/modules/price-router/PriceRouter.sol
crispymangoes: Gravity bridge

crispymangoes: So Somm Governance basically

src/modules/price-router/PriceRouter.sol
crispymangoes: So we want it to be long enough that governance can react if the multisig is acting maliciously, but short enough that if we see a chainlink feed is being deprecated, we can update our pricing info in time before our pricing calls start reverting, so 7 days is a nice middle ground.

test/Registry.t.sol
crispymangoes: Yesm this process, then RYE, and RYGOV cellars will need to call `cachePriceRouter` to use the newest one. Though they dont need to do that until they need to start pricing some exotic asset like balancer lp

test/testPriceRouter/PriceRouter.t.sol
crispymangoes: So this test is useful because if you are checking the price of LINK and have a USD price feed for it, you are only reading the USD price feed. But if you are checking the price of BARN, which only has an ETH price feed, you are reading the BARN price feed, and the ETH-USD price feed. 

We already know through other tests that the main price feed(IE BARN) is checked, but this check confirms that we are also validating the ETH-USD price feed is good.

test/testPriceRouter/PriceRouter.t.sol
crispymangoes: Because if we do the real deal, then the test needs to have its block timestamp fast forwarded by 7 days, which causes all the chainlink price feed checks to revert bc most of them use a 1 day staleness check. We could get away with doing a true happy path test, if we created mock price feeds, but given the audit is basically starting, I dont think it is worth the extra time.

test/testPriceRouter/PriceRouter.t.sol
crispymangoes: Not really, this test is verifying that our math in getExchangeRate is correct, so it is independent from actually seeing if how our derivative is calculating price is correct.

test/testPriceRouter/PriceRouter.t.sol
crispymangoes: So 95% of the pricing logic for TWAPs is coming from UniV3s oracle library, so checking that a TWAP asset price is returning something reasonable is sufficient to see if we set it up correctly.

crispymangoes: I added an RPL case to the exchangeRate test

0xEinCodes (changes requested):
Great job mate. Lots of stuff here. I've gone through most of the PR and left comments (which I see you've responded too a bit this weekend :) ). We'll continue the conversation there and async via chats.

I left the Balancer and Curve related work to offline conversations since there are a few aspects being ironed out on them.

0xEinCodes (approved):

src/modules/price-router/Extensions/BalancerLinearPoolExtension.sol
```diff
@@ -0,0 +1,86 @@
...
+    IVault public immutable balancerVault;
+
+    constructor(PriceRouter _priceRouter, IVault _balancerVault) Extension(_priceRouter) {
+        balancerVault = _balancerVault;
+    }
+
+    /**
+     * @notice Aave Derivative Storage
+     */
+    mapping(ERC20 => bytes32) public getBalancerWeightedPoolDerivativeStorage;
```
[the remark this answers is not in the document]
0xEinCodes: This has been removed.

src/modules/price-router/Extensions/BalancerLinearPoolExtension.sol
```diff
@@ -0,0 +1,86 @@
...
+        ERC20 asset,
+        PriceRouter.PriceCache[PRICE_CACHE_SIZE] memory cache
+    ) external view override returns (uint256) {
+        _ensureNotInVaultContext(balancerVault);
+        IBalancerPool pool = IBalancerPool(address(asset));
+
+        ERC20 mainToken = ERC20(pool.getMainToken());
+
+        uint256 priceBpt = priceRouter.extensionGetPriceInUSD(mainToken, cache).mulDivDown(
+            pool.getRate(),
```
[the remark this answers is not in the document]
0xEinCodes: This has been removed

src/modules/price-router/Extensions/Extension.sol
```diff
@@ -0,0 +1,29 @@
...
+
+import { ERC20 } from "src/base/ERC4626.sol";
+import { PriceRouter } from "src/modules/price-router/PriceRouter.sol";
+import { Math } from "src/utils/Math.sol";
+
+abstract contract Extension {
+    uint8 public constant PRICE_CACHE_SIZE = 8;
+
+    modifier onlyPriceRouter() {
+        if (msg.sender != address(priceRouter)) revert("Only the price router can call this");
```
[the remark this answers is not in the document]
0xEinCodes: Fixed!

src/modules/price-router/PriceRouter.sol
```diff
@@ -120,18 +124,20 @@ contract PriceRouter is Ownable, AutomationCompatibleInterface {
         uint256 _expectedAnswer
     ) external onlyOwner {
         if (address(_asset) == address(0)) revert PriceRouter__InvalidAsset(address(_asset));
+
+        // TODO
+        // if (getAssetSettings[_asset].derivative > 0) revert("Asset already added");
+
         // Zero is an invalid derivative.
         if (_settings.derivative == 0) revert PriceRouter__UnkownDerivative(_settings.derivative);
```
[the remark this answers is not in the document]
0xEinCodes: Fixed

src/modules/price-router/PriceRouter.sol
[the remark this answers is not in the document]
0xEinCodes: Discussed and resolved, bad sources will result in Cellars being paused essentially while new sources are linked to.

src/modules/price-router/PriceRouter.sol
```diff
@@ -146,6 +152,8 @@ contract PriceRouter is Ownable, AutomationCompatibleInterface {
         emit AddAsset(address(_asset));
     }
 
+    function editAsset() external onlyOwner {}
```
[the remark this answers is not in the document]
0xEinCodes: My comment is outdated. Moving on!

src/modules/price-router/PriceRouter.sol
0xEinCodes: We have discussed this, sounds good!

src/modules/price-router/PriceRouter.sol
0xEinCodes: Right, the timeline is (soon):

- `editAsset` would occur, let's say it's malicious.
- Community sees it, and enacts a governance action which requires 2 days minimum. This could include arbitrary calls in the governance action itself, but it could be done after the vote has gone through if need be. 
- This gives enough time for it.

test/Registry.t.sol
0xEinCodes: We discussed this, and the opt-in required from the Cellars makes sense with this process.

test/testPriceRouter/PriceRouter.t.sol
0xEinCodes: We'll leave comment open, the test is a nice to have where we check different derivatives, although we do check them in an isolated mode through other tests.

This was discussed today.

test/testPriceRouter/PriceRouter.t.sol
0xEinCodes: It sounds like moving forward with the mock set up for a true happy path test in the future (aka closer to when BPT cellars or other cellars requiring `PriceRouterV2` is needed) would be good. Let's open an issue on this and address later.

0xEinCodes:
We went through this PR on a call, and resolved any open discussion points. We are good to go! Nice work!

PR #112, opened 2023-05-09
Write `BalancerPoolAdaptor.sol` and `BalancerPoolAdaptor.t.sol`

0xEinCodes:
## Core changes:

- [x] Stub out code w/ referenced docs on how to interact with specific pool types in join/exit functions
   - [/] Understand how the relayer works, I think that’s indicative of the types of functions we will need
- [x] Stub out strategist functions for joins/exits/swaps for respective pools
   - Weighted, Stables & Linear Pools require joins/exits
   - Composable Stable Pools require swaps
   - **NOTE: These are all encompassed into using the `BalancerRelayer` and the complexities of interacting with different types of the Relayer are out of scope for the `BalancerPoolAdaptor` development. Details of how to encode calldata for different actions for Strategists will be outlined as per Issue #113**
- [x] Peer review w/ Crispy on adaptor design/approach
- [x] Revise implementation code
- [ ] Write unit tests & integration tests for Scope 1 (see Notes section below)
- [ ] Peer review & Finalize
- [ ] Final Peer review & Finalize for audit submission

## Notes:
- Managed Pools outside of scope since they are not fully implemented/in use in Balancer ecosystem yet (to be done in a later adaptor)

There are a number of different phases / scopes for the `BalancerPoolAdaptor` development. They are as follows:
- Scope 1: Setup and properly calling relayer with basic `joinData` (using mockChainlinkDataFeed for bb-a-usd : USDC for tests)
- Scope 2: Integrate fully functioning PriceRouterV2 and Balancer specific `Extension` contract
- Scope 3: Source and prepare docs or other resources for strategists to have calldata at the ready for use with `BalancerPoolAdaptor.sol`

## Live Discussion Topics (discussed during the Draft PR stage)
- Strategist `adaptorData` will need to be carefully input. A new issue, issue #113 will be used to record best practices for this. Updates to Balancer will have to be tracked in case there are breaking changes (version updates, etc.).

## References:
1. Balancer Docs for indication of what pools use joins/exits, vs swaps: 
2. Relayer details:
3. Docs on Relayer:

<details markdown='1'><summary> ℹ️ Detailed Notes on DD </summary>

This documents some of the distilled steps taken so far in the due diligence/development of this adaptor:

1. Relayer usage:
- Initially thought as something only EOAs were to use, the `BalancerRelayer` was explored more via assessing `VaultAction.sol`, `BalancerRelayer.sol`, etc. As well as a thorough discussion with CrispyMangoes.
- Initial singular functions were created for join, exit, and swaps, but that was deemed unnecessary if `BalancerRelayer.sol` was to be used.
2. Gauge Interactions:
- Mainly researched Balancer gauge codebase, docs, and Curve docs and codebase for gauge details too.
- Some interfaces are not readily available so some needed to be created. The interfaces I'm referring to need work still but are just roughly done for the first commit.
- `msg.sender` within gauge internal accounting needs to be confirmed to be `cellar` instead of `gravityBridge` addresses when dealing with depositing and withdrawing `bpts`

</details>

<!--- example - Design decisions made in PR>
<!--- example - Remember running `forge fmt` before committing the PR>
<!--- example - Reference for how to run relevant tests i.e. forge test --match-contract X --fork-url $RPC_URL -vvv>


0xEinCodes:
Starting the Draft PR now, even though the rough draft implementation code is all that is complete (tests && compilation yet to be done) because it is best to start getting feedback on the design on Github early on. 

Please feel free to add comments, or even revised commits at your leisure. I will continue working on the outlined TODOs at the beginning of this Draft PR. I will move this to a formal PR once I have the implementation code complete and unit tests working.

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
```diff
@@ -0,0 +1,295 @@
...
+ * @notice Allows Cellars to interact with Weighted, Stable, and Linear Balancer Pools (BPs).
+ * @author 0xEinCodes and CrispyMangoes
+ * TODO: This contract is still a WIP, Still need to go through TODOs and resolve relevant github issues
+ * TODO: Add event emissions where necessary
+ */
+contract BalancerPoolAdaptor is BaseAdaptor {
+    using SafeTransferLib for ERC20;
+
+    //==================== Adaptor Data Specification ====================
+    // adaptorData = abi.encode(ERC20 _bpt, address _poolId, JoinPoolRequest _request)
```
crispymangoes, line 24: I think you can get away with just having the bpt address as the adaptor data

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
```diff
@@ -0,0 +1,295 @@
...
+    function depositBPT(address _poolGauge, uint256 _amountIn, bool _claim_rewards) external {
+        ILiquidityGaugev3Custom poolGauge = ILiquidityGaugev3Custom(_poolGauge);
+        IERC20 bpt = IERC20(poolGauge.lp_token());
+
+        if (bpt.balanceOf(address(this)) <= _amountIn) {
+            revert BalancerPoolAdaptor__NotEnoughToDeposit();
+        }
+        uint256 amountStakedBefore = poolGauge.balanceOf(address(this));
+
+        uint256 amountIn = _maxAvailable(address(bpt), _amountIn);
```
crispymangoes, line 226: <3 this

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
```diff
@@ -0,0 +1,295 @@
...
+     * @param _claim_rewards whether or not to claim pending rewards too (true == claim)
+     * @dev 1.) Interface custom as Balancer/Curve do not provide for liquidityGauges.
+     * @dev 2.) Strategists to provide _poolGauge (which they can get via off-chain querying methods)
+     * TODO: Finalize interface details when beginning to do unit testing
+     */
+    function depositBPT(address _poolGauge, uint256 _amountIn, bool _claim_rewards) external {
+        ILiquidityGaugev3Custom poolGauge = ILiquidityGaugev3Custom(_poolGauge);
+        IERC20 bpt = IERC20(poolGauge.lp_token());
+
+        if (bpt.balanceOf(address(this)) <= _amountIn) {
```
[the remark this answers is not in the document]
crispymangoes: Oh also this logic will revert if `_maxAvailable` comes after this line. So either remove this check, or move `_maxAvailable` above this line

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
```diff
@@ -0,0 +1,295 @@
...
+            revert BalancerPoolAdaptor__NotEnoughToDeposit();
+        }
+        uint256 amountStakedBefore = poolGauge.balanceOf(address(this));
+
+        uint256 amountIn = _maxAvailable(address(bpt), _amountIn);
+        bpt.approve(_poolGauge, amountIn);
+        poolGauge.deposit(amountIn, address(this), _claim_rewards); // address(this) is cellar address bc delegateCall
+        uint256 actualAmountStaked = poolGauge.balanceOf(address(this)) - amountStakedBefore;
+        if (actualAmountStaked >= amountIn) {
```
crispymangoes, line 230: I dont think we need this check in here

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
```diff
@@ -0,0 +1,295 @@
...
+     * @param _poolGauge address of rewards gauge
+     * @param _amountOut number of bpts to withdraw
+     * @param _claim_rewards whether or not to claim pending rewards too (true == claim)
+     * @dev 1.) Interface custom as Balancer/Curve do not provide for liquidityGauges.
+     * @dev 2.) Strategists to provide _poolGauge (which they can get via off-chain querying methods)
+     */
+    function withdrawBPT(address _poolGauge, uint256 _amountOut, bool _claim_rewards) external {
+        ILiquidityGaugev3Custom poolGauge = ILiquidityGaugev3Custom(_poolGauge);
+
+        if (poolGauge.balanceOf(address(this)) >= _amountOut) {
```
crispymangoes, line 248: Again I dont think this check is really needed

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
```diff
@@ -0,0 +1,295 @@
...
+
+    /**
+     * @notice withdraw (unstake) BPT from respective pool gauge
+     * @param _poolGauge address of rewards gauge
+     * @param _amountOut number of bpts to withdraw
+     * @param _claim_rewards whether or not to claim pending rewards too (true == claim)
+     * @dev 1.) Interface custom as Balancer/Curve do not provide for liquidityGauges.
+     * @dev 2.) Strategists to provide _poolGauge (which they can get via off-chain querying methods)
+     */
+    function withdrawBPT(address _poolGauge, uint256 _amountOut, bool _claim_rewards) external {
```
crispymangoes, line 245: Could you add some `_maxAvailable` style logic to this, so if `_amountOut == type(uint256).max` then it sets `_amountOut = poolGuage.balanceOf(address(this));`

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
```diff
@@ -0,0 +1,295 @@
...
+        } else {
+            amountOut = _amountOut;
+        }
+
+        bpt.approve(_poolGauge, amountOut);
+        poolGauge.withdraw(amountOut, _claim_rewards); // msg.sender should be cellar address bc delegateCall. TODO: see issue # <> and confirm address.
+
+        uint256 actualWithdrawn = bpt.balanceOf(address(this)) - unstakedBPTBefore;
+
+        if (actualWithdrawn <= _amountOut) {
```
crispymangoes, line 267: I dont think we need this check either

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
```diff
@@ -0,0 +1,295 @@
...
+    /**
+     * @notice deposit (stake) BPT into respective pool gauge
+     * @param _poolGauge address of rewards gauge
+     * @param _amountIn number of bpts to deposit
+     * @param _claim_rewards whether or not to claim pending rewards too (true == claim)
+     * @dev 1.) Interface custom as Balancer/Curve do not provide for liquidityGauges.
+     * @dev 2.) Strategists to provide _poolGauge (which they can get via off-chain querying methods)
+     * TODO: Finalize interface details when beginning to do unit testing
+     */
+    function depositBPT(address _poolGauge, uint256 _amountIn, bool _claim_rewards) external {
```
crispymangoes, line 217: For all the strategist functions that use `_poolGuage` can they be changed so that they accept a bpt, then use that to query the pool guage address from the pool guage factory?

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
```diff
@@ -0,0 +1,295 @@
...
+     * @dev rewards are only accrue for staked positions
+     * @param _poolGauge address of rewards gauge
+     * @param _rewardToken address of reward token, if not $BAL, that strategist is claiming
+     * TODO: make all verbose text here into github issues
+     * TODO: add checks throughout function
+     */
+    function claimRewards(address _poolGauge, address _rewardToken) public {
+        ILiquidityGaugev3Custom poolGauge = ILiquidityGaugev3Custom(_poolGauge);
+
+        // TODO: checks - though, I'm not sure we need these. If cellar calls `claim_rewards()` and there's no rewards for them then... there are no explicit reverts in the codebase but I assume it reverts. Need to test it though: https://github.com/balancer/balancer-v2-monorepo/blob/master/pkg/liquidity-mining/contracts/gauges/ethereum/LiquidityGaugeV5.vy#L440-L450:~:text=if%20total_claimable%20%3E%200%3A
```
crispymangoes, line 286: Yeah I agree, I dont think we really need this check. The only times I can think when a strategist would be claiming "zero" rewards is
-They call `claimRewards` when they have nothing staked
-They call `claimRewards` twice in the same TX
-They deposit with the claim rewards bool set to true, then call `claimRewards` after

All these cases are user error so I think it is fine to handle this error on the user side(teach the strategist to not claim rewards multiple times during a call)

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
```diff
@@ -0,0 +1,295 @@
...
+     * @notice User deposits are NOT allowed into this position.
+     */
+    function deposit(uint256, bytes memory, bytes memory) public pure override {
+        revert BaseAdaptor__UserDepositsNotAllowed();
+    }
+
+    /**
+     * @notice User withdraws are NOT allowed from this position.
+     */
+    function withdraw(uint256, address, bytes memory, bytes memory) public pure override {
```
crispymangoes, line 125: If you want to try to tackle this feel free, otherwise we can do it together when I am back, but ideally we can allow user withdraws from balancer positions.

To support this, we'll need to implement `withdraw` and `withdrawableFrom`

`withdrawableFrom` will be basically returning the same thing as `balanceOf`.

`withdraw` will need to run a `blockExternalRecevier` check(see other adaptors, like Aave A Token), then if the amount is greater than the amount of BPT the cellar is currently holding in its wallet, it will need to withdraw some BPTs from the guage, then transfer the bpts to the user address provided. Also probs want to have the withdraw NOT harvest rewards since it will add extra gas to the users withdraw TX, and its not really benefitting the strategist much cuz they will probs be calling claim rewards again when rebalancing

crispymangoes (changes requested):
Poggers job my guy, keep up the good work :)

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
0xEinCodes: Ahhh I see. I was getting `adaptorCalls` mixed up with `adaptorData` that is added to the `Registry`.

Is `assetOf()` supposed to only return a single `ERC20` always? If so, what `ERC20` will it be for entering `BoostedPools` or other pools consisting of nested Balancer Pools? IIRC the `_accounting` from `Cellar.sol` will go through all `debt` and `credit` positions and use the `_assetOf()` to get the `ERC20` to use for pricing. This would be custom to the respective adaptor, `BalancerPoolAdaptor.sol` in this case.

I have to double check how the `Extension` design is working, but does it break down yield bearing tokens to base-constituents? AKA can it receive a boosted pool bpt and break it down to 1 ERC20 token (like DAI from bb-a-USD)?

0xEinCodes: For now I'll make it just the bpt address, but I think we need to discuss what I was talking about above more / your expertise would be great :)

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
```diff
@@ -0,0 +1,295 @@
...
+     * TODO: rename params and possibly include others for `exit()` function. ex.) bptOut needs to be renamed to be agnostic.
+     * TODO: see issue for strategist callData
+     */
+    function useRelayer(
+        ERC20[] memory tokensIn,
+        uint256[] memory amountsIn,
+        ERC20 bptOut,
+        bytes[] memory callData
+    ) public {
+        for (uint256 i; i < tokensIn.length; ++i) tokensIn[i].approve(relayer(), amountsIn[i]); // TODO: this may be obsolete based on approval steps necessary for relayer functionality (see TODO below)
```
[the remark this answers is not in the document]
0xEinCodes: Hmm, I think we only need to approve the relayer because it would take our tokens and act on our behalf on the vault right? 

Referencing the docs, it states the relayer is the single point of entry [here](https://docs.balancer.fi/concepts/advanced/relayers.html#how-it-works:~:text=which%20is%20the%20single%20point%20of%20entry%20via%20the%20multicall%20function), and that a user has to apprve each relayer before use [here](https://docs.balancer.fi/concepts/advanced/relayers.html#contracts:~:text=A%20user%20has%20to%20approve%20each%20Relayer%20before%20they%20can%20use%20it.%20To%20check%20if%20a%20Relayer%20is%20approved%20the%20hasApprovedRelayer%20on%20the%20Vault%20can%20be%20used.)

From a quick skim on the code, it looks like the `Relayer` calls `multiCall` which acts upon the `Vault` where things like `joinPool()` could be called. Going down that path more, `joinPool()` shows that transfers eventually happen from `AssetTransfersHandler.sol` (Vault inherits this contract). It is a `safeTransferFrom()` / typical transfer calls. So the `sender` mentioned in this call, is I think the original caller - so the `Cellar` in our case. If we call `relayerApproval` then we are giving `allowance` to the relayer to spend on the Cellar behalf. 

So here's a question with all that: is `msg.sender` the `Cellar` address if we call it using `delegateCall()` from the `BalancerPoolAdaptor` in this eventual function call in `BaseRelayerLibrary.sol`? I think that it is. So we'd be giving approval to the relayer to spend the `Cellar` assets. At the end we'd have to revoke approval for the Relayer on behalf of the Cellar, so change this [value](https://github.com/balancer/balancer-v2-monorepo/blob/21e028ca8fe31beff6c90e007739b85ce8d8c4c6/pkg/vault/contracts/VaultAuthorization.sol#LL60C5-L60C76). I'm not sure how to do this because it doesn't show a way to do so in the aforementioned Balancer contracts. 

```
function setRelayerApproval(
        address relayer,
        bool approved,
        bytes calldata authorisation
    ) external payable {
        require(relayer == address(this) || !approved, "Relayer can only approve itself");
        bytes memory data = abi.encodePacked(
            abi.encodeWithSelector(_vault.setRelayerApproval.selector, msg.sender, relayer, approved),
            authorisation
        );
```

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
```diff
@@ -0,0 +1,295 @@
...
+
+        uint256 startingBpt = BPTout.balanceOf(address(this));
+
+        // TODO: implement proper way of giving Relayer approval; Relayer.setRelayerApproval() or Relayer.approveVault().
```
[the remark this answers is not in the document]
0xEinCodes: I think you're right but I have questions about the revokeApproval aspects. Please see my message before this on your other comment.

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
```diff
@@ -0,0 +1,295 @@
...
+
+        relayer().multicall(callData);
+
+        uint256 endingBpt = BPTout.balanceOf(address(this));
+
+        uint256 amountBptOut = endingBpt - startingBpt;
+
+        uint256 amountBptIn = priceRouter.getValues(tokensIn, amountsIn, bptOut); // TODO: not 100% sure how this math all works out. I think this roughly calcs the BPTValue In, it depends on the source of its priceData cause it has to know how much slippage to account for.
```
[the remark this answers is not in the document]
0xEinCodes: Sweet! Yeah I think there's just general questions I have with how the PriceRouter Extension for Balancer will work in general. We can discuss more!

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
```diff
@@ -0,0 +1,295 @@
...
+
+        // check value in vs value out
+        if (amountBptOut < amountBptIn.mulDivDown(0.9e4, 1e4)) revert("Slippage");
+
+        // revoke token in approval
+        for (uint256 i; i < tokensIn.length; ++i) _revokeExternalApproval(tokensIn[i], relayer());
```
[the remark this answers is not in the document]
0xEinCodes: I'm still not sure about this. See my longer response in your other comment before. If we can't change this variable then we'll have to consider other methods (outlined [here](https://github.com/balancer/balancer-v2-monorepo/blob/21e028ca8fe31beff6c90e007739b85ce8d8c4c6/pkg/vault/contracts/VaultAuthorization.sol#L60)):

`mapping(address => mapping(address => bool)) private _approvedRelayers;`

Perhaps we have our own approval marker within the `Registry.sol` or elsewhere (just brainstorming quick), that we check before engaging with the `Relayer` through the `BalancerPoolAdaptor.sol`?

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
```diff
@@ -0,0 +1,295 @@
...
+        relayer().multicall(callData);
+
+        uint256 endingBpt = BPTout.balanceOf(address(this));
+
+        uint256 amountBptOut = endingBpt - startingBpt;
+
+        uint256 amountBptIn = priceRouter.getValues(tokensIn, amountsIn, bptOut); // TODO: not 100% sure how this math all works out. I think this roughly calcs the BPTValue In, it depends on the source of its priceData cause it has to know how much slippage to account for.
+
+        // check value in vs value out
+        if (amountBptOut < amountBptIn.mulDivDown(0.9e4, 1e4)) revert("Slippage");
```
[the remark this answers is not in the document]
0xEinCodes: I'll have to think about this more. Let's use the value from `BaseAdaptor` for now! Thanks!

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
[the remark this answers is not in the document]
0xEinCodes: I've moved the `_maxAvailable` line up and deleted the `check`

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
0xEinCodes: Yeah I think I did this in the wrong way anyways. We'd be more interested if the `amountStaked` was less than the amount we put in right? I'll remove the error though I wonder if we should still check that the `actualAmountStaked` was at least the amount we put in or a certain amount of it.

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
0xEinCodes: Agreed!

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
0xEinCodes: Good call, added it!

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
0xEinCodes: Hmm interesting, how do we make sure we are getting at least a minimum amount of the requested withdrawn `_amountOut`?

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
0xEinCodes: I will look into that, that sounds like an easier UX for strategists I assume?

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
0xEinCodes: Let's discuss this one further. I'll leave the comment for now during development.

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
0xEinCodes: Kk I'll look into this as well and update this PR with a message once it is done. The thing I'm wondering about though is whether or not we want users to be able to withdraw directly from this `Adaptor` contract. They should be able to call `redeem()` or `withdraw()` from the `Cellar` right? That will do the functionality that you mention I thought.

0xEinCodes:
Just responded to most of your comments. I've also pushed a commit with some of the change requests. There are some messy comments in there though that you can disregard (in the code) - I'm just running out the door but will come back and make the other changes and also tackle test writing throughout the weekend / early this coming week!

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
0xEinCodes: I've made the changes!

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
0xEinCodes: I've removed it but the question still stands :)

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
0xEinCodes: `depositBPT()`, `withdrawBPT()`, and `claimRewards()` are changed so they have an input `param` as the `address _bpt` now.

For `depositBPT()` and `withdrawBPT()` it should make sense! `claimRewards()` is up in the air as outlined in our other conversation in this PR.

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
crispymangoes: The extension does break down yield bearing tokens into their underlyings in order to find a USD value of BPTs

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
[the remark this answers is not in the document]
crispymangoes:
```
function setRelayerApproval(bool state) public {
     vault().setRelayerApproval(address(relayer(), state);
}
```

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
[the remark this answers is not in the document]
0xEinCodes: As per our chat we need to do the following:
- Check how often the relayer is actually changed (if it's < 1 year, then we'll have to consider setting this adaptor up so it changes with it, versus redeploying a new adaptor which can be cost-intensive)
- If it is < 1 year, a method that we could use (and may use regardless), is to do the following:
   - Something along the lines of
   ```
   address currentRelayer = getRelayerFromBalancer();
   if (currentRelayer != relayer()) {
      // check if new relayer is approved by Balancer
   }
   ```

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
crispymangoes: Is pool guage upgradeable? If so then we can keep this check, otherwise the balanceOf logic is pretty simple so this check is not needed

0xEinCodes:
## Ongoing Notes Keeping Track of What's Remaining TODO:
- [ ] CHECK `IChainlinkAggregator` interface and create the `DUMMY_BB_A_USD_PRICEFEED` abiding by it. In it, have a setter and a getter for the price with respect to the required function calls: `latestRoundData()` --> **This is done to temporarily bypass Cellar checks since that is not the scope of the `BalancerPoolAdaptor` tests right now**
   - NOTE: `PriceRouterV2` and `BalancerExtensions` for pricing sources will be checked in later phases of `BalancerPoolAdaptor.sol` development.
- [ ] Write `setup()` to incorporate `DUMMY_BB_A_USD_PRICEFEED` and troubleshoot to have it compile.
- [ ] Obtain `Cellar` address in test and manually create `joinData` to use for first phase of testing.
- [ ] Create Happy Path tests
- [ ] Create Non-Happy Path tests

0xEinCodes:
Tests have been progressing and are currently in another temporary branch named "crispy-balancer." I will migrate changes over to this branch later, or merge them. Next steps are to work on include (as per convos with Crispy):

- Pool Gauge as Params: Add poolGauge as a parameter within adaptorData
    - Change `BalancerPoolAdaptor.sol` accordingly
    - Change tests accordingly
- BalanceOf test: Now that gauge is properly working, make sure that it checks the gauges properly for each BPT that the cellar manages / has managed?
- Exit BPTs Function: Make a new function that looks similar to `UseRelayer()` to allow Cellar’s to exit BPTs
- withdraw() Function: Make the implementation code for this new function in `BalancerPoolAdaptor.sol` where users can get some or all of their BPTs if they are trying to exit the cellar but the cellar is already in a wound-up position (aka a BPT, or staked BPT) —> if staked, unstake it and send resultant BPT?
- Basic tests for deposit into and withdraw from gauge: 
    - Deposit:
        - FullDeposit: Check balance of staked BPT ERC20 coupons you get back
        - Check that balanceOf() accounts for staked BPTs
        - PartialDeposit(): Check balance of BPT you get input and the leftover in your account
    - Withdraw:
        - FullWithdraw(): Check balance of BPT you get back
        - PartialWithdraw(): Check balance of BPT you get back and the leftover
        - Carry out a general withdraw() where “stakedBPT” needs to be unstaked
- BIG CHECK:  Check if the relayer allows the approval for approval of another user to manage your funds? (Ask Crispy about this, it’s something he mentioned when he got the exit data)

src/modules/adaptors/Balancer/BalancerPoolAdaptor.sol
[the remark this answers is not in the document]
0xEinCodes: I think I clarified this for myself. Relayer is a proxy, basically outlined [here](https://docs.balancer.fi/sdk/technical-reference/relayer.html#:~:text=Relayers%20are%20(user%20opt%2Din%2C%20audited)%20contracts%20that%20can%20make%20calls%20to%20the%20vault%20(with%20the%20transaction%20%E2%80%9Csender%E2%80%9D%20being%20any%20arbitrary%20address)%20and%20use%20the%20sender%E2%80%99s%20ERC20%20vault%20allowance%2C%20internal%20balance%20or%20BPTs%20on%20their%20behalf.). So giving the `vault` the approval and revoking it at the end of the tx is right.

0xEinCodes:
This PR is being migrated to a new PR #126.

Recent comments left with 👀 are to be addressed in the new PR amongst other TODOs. This PR is closed due to some errors and being shelved and thus accumulating technical debt since the rest of the `main` has been updated with the recent audits.

PR #117, opened 2023-05-30
Grab price router from correct source

crispymangoes:


0xEinCodes (changes requested):
This change, and the repeat ones make sense to me as you are upgrading the adaptors to be consistent with others in the protocol. Although, [from doing a quick check in the other adaptors](https://github.com/search?q=repo%3APeggyJV%2Fcellar-contracts%20path%3Asrc%2Fmodules%2Fadaptors%2F%20priceRouter&type=code), it looks like the only ones that also have this call layout is: OneInch and ZeroX adaptors. In them, it looks like they are querying `address(this)` when I think it should be like the adjustments you made here and be `msg.sender`. Please confirm.

crispymangoes:
Yeah so zerox and 1inch use address(this) because those functions are being called in a delegate call context, where the a token adaptors, and uniswap V3 adaptor changes are being called in a normal function context, so the cellar address is msg.sender

0xEinCodes:
> Yeah so zerox and 1inch use address(this) because those functions are being called in a delegate call context, where the a token adaptors, and uniswap V3 adaptor changes are being called in a normal function context, so the cellar address is msg.sender

Ah, yeah that makes sense! We discussed this offline, and I just didn't poke into the code I was linking to in my initial comment. I can see that they were not mutative explaining the misinterpretation of the context. Looks good to me then!

0xEinCodes (approved):
See my last comment. Nice!

PR #118, opened 2023-06-02
Feat/axelar proxy

crispymangoes:


src/AxelarProxy.sol
```diff
@@ -0,0 +1,67 @@
...
+     */
+    function _execute(string calldata sourceChain, string calldata, bytes calldata payload) internal override {
+        // Make sure executions are still allowed.
+        if (stopExecute) revert AxelarProxy__ExecutionStopped();
+
+        // Validate Source Chain
+        if (keccak256(bytes(sourceChain)) != SOMMELIER_CHAIN_HASH) revert AxelarProxy__WrongSource();
+
+        // Execute function call.
+        (address target, bytes memory callData) = abi.decode(payload, (address, bytes));
```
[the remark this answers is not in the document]
0xEinCodes: Also I think the `callData` here is the `payload` right? So having `_payload` as the initial param in the first place might be a solution.

src/AxelarProxy.sol
```diff
@@ -0,0 +1,47 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.16;
+
+import { AxelarExecutable } from "lib/axelar-gmp-sdk-solidity/contracts/executable/AxelarExecutable.sol";
+import { Address } from "@openzeppelin/contracts/utils/Address.sol";
+
```
0xEinCodes, line 6: Can you add a description to the contract outlining how it will work with the Sommelier protocol? In my bigger message at the end of this review, I outline how I perceive your code in this PR is to be carried out, but descriptions would help. It wouldn't hurt for the Auditor too :)

src/AxelarProxy.sol
```diff
@@ -0,0 +1,47 @@
...
+    error AxelarProxy__WrongSource();
+    error AxelarProxy__NoTokens();
+
+    bytes32 public constant SOMMELIER_CHAIN_HASH = keccak256(bytes("sommelier"));
+
+    constructor(address gateway_) AxelarExecutable(gateway_) {}
+
+    /**
+     * @notice Execution logic.
+     * @dev Verifies message is from Sommelier, otherwise reverts.
```
0xEinCodes, line 21: Please add note about how the `_execute()` function checks with the `Gateway` as well from the `Source` chain(?). Please confirm if that is how it works too, this was just from separate discussions with the protocol team, although I may have interpreted it incorrectly.

src/AxelarProxy.sol
```diff
@@ -0,0 +1,47 @@
...
+        // Execute function call.
+        (address target, bytes memory callData) = abi.decode(payload, (address, bytes));
+        target.functionCall(callData);
+
+        emit LogicCallEvent(target, callData);
+    }
+
+    /**
+     * @notice Execution with token logic.
+     * @dev Not supported.
```
0xEinCodes, line 36: perhaps add a bit more here to the dev note, or on the @notice.

Ex.) "@notice Execution with token logic as per Axelar functionality. Not supported in this contract."

src/mocks/MockSommelier.sol
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.16;
+
+import { Address } from "@openzeppelin/contracts/utils/Address.sol";
+import { ERC20 } from "src/base/Cellar.sol";
+import { IAxelarGateway } from "lib/axelar-gmp-sdk-solidity/contracts/interfaces/IAxelarGateway.sol";
+import { IAxelarGasService } from "lib/axelar-gmp-sdk-solidity/contracts/interfaces/IAxelarGasService.sol";
+
+contract MockSommelier {
```
0xEinCodes, line 9: Please add a description to this as well. Just to help document things further. I was a bit confused why this was a `MockSommelier` contract working with Arbitrum vs Polygon as per the other couple of files within this PR.

script/test/Polygon/SendMessageToArbitrum.s.sol
```diff
@@ -0,0 +1,41 @@
...
+    address private usdcOnArb = 0xFF970A61A04b1cA14834A43f5dE4533eBDDB5CC8;
+
+    MockSommelier private mockSomm;
+
+    function run() external {
+        vm.startBroadcast();
+
+        mockSomm = new MockSommelier();
+
+        mockSomm.sendMessage{ value: 10 ether }(destAddress, usdcOnArb, devOwner, 777);
```
0xEinCodes, line 37: Is the leftover `gas` just returned I presume?

script/test/Arbitrum/DeployProxy.s.sol
```diff
@@ -0,0 +1,29 @@
...
+import "forge-std/Script.sol";
+import { Math } from "src/utils/Math.sol";
+
+/**
+ * @dev Run
+ *      `source .env && forge script script/test/Arbitrum/DeployProxy.s.sol:DeployProxyScript --rpc-url $ARBITRUM_RPC_URL  --private-key $PRIVATE_KEY —optimize —optimizer-runs 200 --verify --etherscan-api-key $ARBISCAN_KEY --slow --broadcast`
+ * @dev Optionally can change `--with-gas-price` to something more reasonable
+ */
+contract DeployProxyScript is Script {
+    address private devOwner = 0x552acA1343A6383aF32ce1B7c7B1b47959F7ad90;
```
0xEinCodes, line 15: Who will be the owner

src/AxelarProxy.sol
```diff
@@ -0,0 +1,47 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.16;
+
+import { AxelarExecutable } from "lib/axelar-gmp-sdk-solidity/contracts/executable/AxelarExecutable.sol";
+import { Address } from "@openzeppelin/contracts/utils/Address.sol";
+
+contract AxelarProxy is AxelarExecutable {
```
0xEinCodes, line 7: What permissions exist for this proxy now that `Ownable` no longer applies?

0xEinCodes:
In general looks good to move forward to audit. I made some small suggestions and had questions that would be great to get clarity on before wrapping it up. 

That all said, I think there are the risk vectors associated to what the `toggle` was supposed to do in `AxelarProxy.sol`. What is the resolution process if Axelar is hacked? I suppose Cellars fully relying on msg communications using Axelar would be in `withdrawOnly` mode where users could withdraw in those networks. Obviously that is not ideal for users though.

**I think Resolution processes should be discussed openly in an issue as this feature is developed and deployed.**

---
As well, please confirm that I am understanding this PR details (might be just that I am newer to the Axelar setup since we're both working on a multitude of different things nowadays!)

- `AxelarProxy.sol`: the proxy contract itself that will be deployed on alternative chains. It is the touch point for us, and showcases that we can set things up to execute various actions in an alternative chain.
- `DeployProxy.s.sol`: the deployment script just testing out deployment of the Proxy contract.
- `MockSommelier.sol` & `SendMessageToArbitrum.s.sol` —> together tests out how Sommelier (Steward) would actually make the call with Axelar.
    - `MockSommelier.sol`: showcases how Sommelier (Steward) would prepare data and pass it along to Axelar gateway, ultimately to a proxy contract on an alternative chain.
    - `SendMessageToArbitrum.s.sol`: simply the script to carry out this test.
        - Question: Sommelier determines what the payload will carry out, and the destination chain `Proxy` contract will check with the Axelar `Gateway` if a certain `payload` is an approved action right? If so, how does that all work, this is something that needs to be well documented / our resources need to point to it so others can understand this when they want to work with the Sommelier protocol. 
                - On the Sommelier side, it would be regular $SOMM governance dictating what is passed to be created on these alt chains, and then further coordination/whitelisting with Axelar would be needed I assume?

src/AxelarProxy.sol
```diff
@@ -0,0 +1,67 @@
...
+     */
+    function toggleExecution() external onlyOwner {
+        stopExecute = stopExecute ? false : true;
+    }
+
+    /**
+     * @notice Execution logic.
+     * @dev Verifies message is from Sommelier, otherwise reverts.
+     */
+    function _execute(string calldata sourceChain, string calldata, bytes calldata payload) internal override {
```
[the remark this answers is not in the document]
crispymangoes: Oh `calldata` is a solidity keyword describing where this value is stored.

src/AxelarProxy.sol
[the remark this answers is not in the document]
crispymangoes: So `callData` is data you want to pass into the `functionCall` function, where `payload` is the abi encoded target address, and function call data.

script/test/Polygon/SendMessageToArbitrum.s.sol
crispymangoes: Yes, it is, it is refunded on the chain where the message was originally sent. Though we should talk with protocol team this week, and make sure that they dont need to send tokens with their execution calls in order to pay for gas, cuz I am not sure how we will pay for gas on Somm.

src/AxelarProxy.sol
crispymangoes: None, it is completely immutable

src/AxelarProxy.sol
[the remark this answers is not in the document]
0xEinCodes: Makes sense! Nice.

src/AxelarProxy.sol
0xEinCodes: We discussed this, and the `AxelarProxy.sol` `execute()` function has logic that lays this out.

src/mocks/MockSommelier.sol
0xEinCodes: You've added this! Nice!

script/test/Polygon/SendMessageToArbitrum.s.sol
0xEinCodes: #119 will continue the convo.

src/AxelarProxy.sol
0xEinCodes: Yep sounds good.

PR #120, opened 2023-06-05
Feat/morpho

crispymangoes:


src/modules/adaptors/Morpho/MorphoAaveV3DebtTokenAdaptor.sol
```diff
@@ -0,0 +1,156 @@
...
+        bytes32 positionHash = keccak256(abi.encode(identifier(), true, abi.encode(underlying)));
+        uint32 positionId = Cellar(address(this)).registry().getPositionHashToPositionId(positionHash);
+        if (!Cellar(address(this)).isPositionUsed(positionId))
+            revert MorphoAaveV3DebtTokenAdaptor__DebtPositionsMustBeTracked(underlying);
+
+        // Borrow from morpho.
+        morpho().borrow(underlying, amountToBorrow, address(this), address(this), maxIterations);
+
+        // Check that health factor is above adaptor minimum.
+        uint256 healthFactor = _getUserHealthFactor(address(this));
```
0xEinCodes, line 127: Interesting, does Morpho contracts make it so that you cannot check the health factor as a preview prior to executing `morpho.borrow()`? If they do provide it, then moving this before the `borrow()` would be more gas efficient no?

src/modules/adaptors/Morpho/MorphoAaveV3DebtTokenAdaptor.sol
```diff
@@ -0,0 +1,156 @@
...
+    function morpho() internal pure returns (IMorpho) {
+        return IMorpho(0x33333aea097c193e66081E930c33020272b33333);
+    }
+
+    /**
+     * @notice Minimum Health Factor enforced after every borrow.
+     * @notice Overwrites strategist set minimums if they are lower.
+     */
+    function HFMIN() internal pure returns (uint256) {
+        return 1.05e18;
```
0xEinCodes, line 58: Should this be adjustable with proper permissions per Cellar (and perhaps a timelock)? The reason I ask is because I am wondering about scenarios where the health factor for Morpho or the underlying protocol (AAVE) allows situations with lower HF. That said, 1.05 HF is pretty low, so maybe better to leave it as is.

src/interfaces/external/Morpho/IMorpho.sol
```diff
@@ -0,0 +1,90 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.16;
+
+interface IMorpho {
```
0xEinCodes, line 4: These don't seem to line up completely with the function sigs shown in morpho repo, but I could be mistaken. Please clarify. Why not have two separate Morpho interfaces? (AAVE v2 - blob/main/src/aave-v2/interfaces/IMorpho.sol & AAVE v3 - https://github.com/morpho-org/morpho-aave-v3/blob/main/src/interfaces/IMorpho.sol)

src/modules/adaptors/Morpho/MorphoAaveV2ATokenAdaptor.sol
```diff
@@ -0,0 +1,171 @@
...
+        address receiver,
+        bytes memory adaptorData,
+        bytes memory configData
+    ) public override {
+        // Run external receiver check.
+        _externalReceiverCheck(receiver);
+
+        // Make sure position is setup to be liquid.
+        bool isLiquid = abi.decode(configData, (bool));
+        if (!isLiquid) revert BaseAdaptor__UserWithdrawsNotAllowed();
```
0xEinCodes, line 86: Doesn't Morpho have the `fallback` setup so it switches out P2P arrangements when one party exits while capital is still deployed? Essentially where another participant is matched, or worst case the underlying pool fills in when a party leaves?

0xEinCodes: Or is this a design intent for the Adaptors here?

src/modules/adaptors/Morpho/MorphoAaveV2ATokenAdaptor.sol
```diff
@@ -0,0 +1,171 @@
...
+
+        IAaveToken aToken = abi.decode(adaptorData, (IAaveToken));
+
+        // Withdraw assets from Morpho.
+        morpho().withdraw(address(aToken), assets, receiver);
+    }
+
+    /**
+     * @notice Uses configurartion data to determine if the position is liquid or not.
```
0xEinCodes, line 95: nit: spelling of "configuration"

src/modules/adaptors/Morpho/MorphoAaveV2ATokenAdaptor.sol
```diff
@@ -0,0 +1,171 @@
...
+     * @param assets the amount of assets to withdraw from Morpho
+     * @param receiver the address to send withdrawn assets to
+     * @param adaptorData adaptor data containing the abi encoded aToken
+     * @param configData abi encoded bool indicating whether the position is liquid or not
+     */
+    function withdraw(
+        uint256 assets,
+        address receiver,
+        bytes memory adaptorData,
+        bytes memory configData
```
0xEinCodes, line 79: where is `configData` coming from?

src/modules/adaptors/Morpho/MorphoAaveV2ATokenAdaptor.sol
```diff
@@ -0,0 +1,171 @@
...
+    /**
+     * @notice Returns the balance in underlying of collateral.
+     * @param poolToken the Aave V2 a Token user has supplied as collateral
+     * @param user the address of the user to query their balance of
+     */
+    function _balanceOfInUnderlying(address poolToken, address user) internal view returns (uint256) {
+        (uint256 inP2P, uint256 onPool) = morpho().supplyBalanceInOf(poolToken, user);
+
+        uint256 balanceInUnderlying;
+        if (inP2P > 0) balanceInUnderlying = inP2P.mulDivDown(morpho().p2pSupplyIndex(poolToken), 1e27);
```
0xEinCodes, line 167: Let's walk through this line.

src/modules/adaptors/Morpho/MorphoAaveV3ATokenCollateralAdaptor.sol
```diff
@@ -0,0 +1,141 @@
...
+        underlying.safeApprove(address(morpho()), assets);
+
+        morpho().supplyCollateral(address(underlying), assets, address(this));
+
+        // Zero out approvals if necessary.
+        _revokeExternalApproval(underlying, address(morpho()));
+    }
+
+    /**
+     @notice Cellars must withdraw from Morpho, check if collateral is backing any loans
```
0xEinCodes, line 61: Same question as posted on v2 adaptors

src/modules/adaptors/Morpho/MorphoAaveV3ATokenP2PAdaptor.sol
```diff
@@ -0,0 +1,154 @@
...
+    // Where:
+    // `underlying` is the ERC20 position this adaptor is working with
+    //================= Configuration Data Specification =================
+    // configurationData = abi.encode(uint256 maxIterations);
+    // Where:
+    // `maxIterations` is some number greater than 0 but less than MAX_ITERATIONS,
+    // allows assets to be P2P matched.
+    //====================================================================
+
+    uint256 internal constant MAX_ITERATIONS = 10;
```
0xEinCodes, line 27: Where did these iteration values come from?

0xEinCodes: Is this in place for when maxGas caps won't necessarily guarantee adequate APR because the Cellar may not match with a P2P APY?

0xEinCodes (changes requested):
Reviewed contracts aside from `RealYieldGov` aspects and full tests (aside from running on local machine). We will discuss further!

src/modules/adaptors/Morpho/MorphoAaveV2ATokenAdaptor.sol
0xEinCodes: This comment can be disregarded after discussing things further. I believe you will be adding in extra comments to help clarify what you mean by the misconfigured config data.

src/modules/adaptors/Morpho/MorphoAaveV2ATokenAdaptor.sol
```diff
@@ -0,0 +1,171 @@
...
+     * @notice Allows strategists to lend assets on Morpho.
+     * @dev Uses `_maxAvailable` helper function, see BaseAdaptor.sol
+     * @param aToken the aToken to lend on Morpho
+     * @param amountToDeposit the amount of `tokenToDeposit` to lend on Morpho.
+     */
+    function depositToAaveV2Morpho(IAaveToken aToken, uint256 amountToDeposit) public {
+        ERC20 underlying = ERC20(aToken.UNDERLYING_ASSET_ADDRESS());
+        amountToDeposit = _maxAvailable(underlying, amountToDeposit);
+        underlying.safeApprove(address(morpho()), amountToDeposit);
+        morpho().supply(address(aToken), amountToDeposit);
```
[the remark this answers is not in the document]
0xEinCodes: Any pooled amount will be matched as new borrowers enter the protocol. So putting it all in, assuming APY is fair enough, is fine.

src/modules/adaptors/Morpho/MorphoAaveV3ATokenCollateralAdaptor.sol
0xEinCodes: Resolved from before!

src/modules/adaptors/Morpho/MorphoAaveV3DebtTokenAdaptor.sol
crispymangoes: Yeah I think the easiest way to see if we can do a borrow is to actually try doing the borrow, then it will either revert in the morpho contracts(if HF is < 1), or in our adaptor if HF is less than HFMIN. 

There isn't a straight forward function we can call to check to see how a new borrow would impact our health factor, and even if there was, we'd end up spending more gas running the calculation everytime we borrowed, vs the gas we would save from reverting early for the odd times where we borrowed too much.

src/modules/adaptors/Morpho/MorphoAaveV3DebtTokenAdaptor.sol
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    /**
+     * @notice Code pulled directly from Morpho Position Manager.
+     * https://etherscan.io/address/0x4592e45e0c5DbEe94a135720cCfF2e4353dAc6De#code
+     */
+    function _getUserHealthFactor(address user) internal view returns (uint256) {
+        IMorpho.LiquidityData memory liquidityData = morpho().liquidityData(user);
+
+        return
```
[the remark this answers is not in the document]
crispymangoes: I think we are good on this, with my last comment, but can expand if you want

src/modules/adaptors/Morpho/MorphoAaveV3ATokenP2PAdaptor.sol
crispymangoes: So in their contracts, they have a comment saying a good iterations value is 4, so I somewhat arbitrarily chose 10 as a max. I don't think we would be getting much value from using a value greater than 4, but the main reason why we have a cap is so that strategists can't gas grief our relayer by using a stupid big iterations value.

src/modules/adaptors/Morpho/MorphoAaveV3ATokenP2PAdaptor.sol
```diff
@@ -0,0 +1,154 @@
...
+        uint256 iterations = abi.decode(configurationData, (uint256));
+
+        // Withdraw assets from Morpho.
+        morpho().withdraw(address(underlying), assets, address(this), receiver, iterations);
+    }
+
+    /**
+     * @notice Returns the p2p balance of the cellar.
+     */
+    function withdrawableFrom(bytes memory adaptorData, bytes memory) public view override returns (uint256) {
```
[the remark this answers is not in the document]
crispymangoes: I think we discussed this on our call, but can elaborate more if you want. 

I'll check and see if there is an easy way to figure out the available assets to withdraw based off the amount of assets supplied.

src/modules/adaptors/Morpho/MorphoAaveV2ATokenAdaptor.sol
```diff
@@ -0,0 +1,175 @@
...
+     * @param poolToken the Aave V2 a Token user has supplied as collateral
+     * @param user the address of the user to query their balance of
+     */
+    function _balanceOfInUnderlying(address poolToken, address user) internal view returns (uint256) {
+        (uint256 inP2P, uint256 onPool) = morpho().supplyBalanceInOf(poolToken, user);
+
+        uint256 balanceInUnderlying;
+        // Morpho indexes are scaled by 27 decimals, so divide by 1e27.
+        if (inP2P > 0) balanceInUnderlying = inP2P.mulDivDown(morpho().p2pSupplyIndex(poolToken), 1e27);
+        if (onPool > 0) balanceInUnderlying += onPool.mulDivDown(morpho().poolIndexes(poolToken).poolSupplyIndex, 1e27);
```
Rubilmax, line 172: Bear in mind that these indexes may be outdated (they are updated at each interaction with the protocol and thus are not guaranteed to be up-to-date). If you want to access the true balance in underlying, you need to use the indexes calculated by the Lens (or calculate them yourself)

src/modules/adaptors/Morpho/MorphoAaveV2DebtTokenAdaptor.sol
```diff
@@ -0,0 +1,147 @@
...
+     * @notice Returns the balance in underlying of debt owed.
+     * @param poolToken the Aave V2 a Token user has debt in
+     * @param user the address of the user to query their debt balance of.
+     */
+    function _balanceOfInUnderlying(address poolToken, address user) internal view returns (uint256) {
+        (uint256 inP2P, uint256 onPool) = morpho().borrowBalanceInOf(poolToken, user);
+
+        uint256 balanceInUnderlying;
+        if (inP2P > 0) balanceInUnderlying = inP2P.mulDivDown(morpho().p2pBorrowIndex(poolToken), 1e27);
+        if (onPool > 0) balanceInUnderlying += onPool.mulDivDown(morpho().poolIndexes(poolToken).poolBorrowIndex, 1e27);
```
Rubilmax, line 144: Bear in mind that these indexes may be outdated (they are updated at each interaction with the protocol and thus are not guaranteed to be up-to-date). If you want to access the true balance in underlying, you need to use the indexes calculated by the Lens (or calculate them yourself)

src/modules/adaptors/Morpho/MorphoAaveV3ATokenCollateralAdaptor.sol
```diff
@@ -0,0 +1,141 @@
...
+    /**
+     * @notice Checks that cellar has no active borrows, and if so reverts.
+     */
+    function withdrawableFrom(bytes memory adaptorData, bytes memory) public view override returns (uint256) {
+        address[] memory borrows = morpho().userBorrows(msg.sender);
+        if (borrows.length > 0) return 0;
+        else {
+            ERC20 underlying = abi.decode(adaptorData, (ERC20));
+            return morpho().collateralBalance(address(underlying), msg.sender);
+        }
```
Rubilmax, line 91:
```suggestion
        else return balanceOf(adaptorData);
```

src/modules/adaptors/Morpho/MorphoAaveV3ATokenCollateralAdaptor.sol
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    /**
+     * @notice Returns the cellars balance of the position in terms of underlying asset.
+     */
+    function balanceOf(bytes memory adaptorData) public view override returns (uint256) {
+        ERC20 underlying = abi.decode(adaptorData, (ERC20));
```
Rubilmax, line 98:
```suggestion
        ERC20 underlying = assetOf(adaptorData);
```

src/modules/adaptors/Morpho/MorphoAaveV2ATokenAdaptor.sol
```diff
@@ -0,0 +1,175 @@
...
+
+    /**
+     * @notice Uses configuration data to determine if the position is liquid or not.
+     */
+    function withdrawableFrom(
+        bytes memory adaptorData,
+        bytes memory configData
+    ) public view override returns (uint256) {
+        bool isLiquid = abi.decode(configData, (bool));
+        if (!isLiquid) return 0;
```
Rubilmax, line 105: There exist a function in Morpho-AaveV2 Lens that can calculate the withdrawable amount: https://github.com/morpho-org/morpho-v1/blob/0df9d1a925d2342ea87f2fc5620d693604d29f3b/src/aave-v2/lens/UsersLens.sol#L53

src/modules/adaptors/Morpho/MorphoAaveV2ATokenAdaptor.sol
crispymangoes: Ah I see I'll take a look at the lens contract, it is okay if these numbers are slightly off, like in our Compound V2 code we use the stored exchange rate mantissa, and that is fine because it never substantially deviates from the actual value. So if that is the same case here, I am not worried about it.

src/modules/adaptors/Morpho/MorphoAaveV2ATokenAdaptor.sol
crispymangoes: Oh this is great! I'll take a look and see if I can convert the math so that I can input a minimum health factor, and get an amount of assets withdrawable that will bring you down to that minimum.

src/modules/adaptors/Morpho/MorphoAaveV3DebtTokenAdaptor.sol
0xEinCodes: Ah fair enough, yeah it's fine-tuning regarding the conditional logic of checking vs the execution call where we borrowed and had to revert.

src/modules/adaptors/Morpho/MorphoAaveV3DebtTokenAdaptor.sol
0xEinCodes: Sounds good, we'll talk about it more but it makes sense.

src/modules/adaptors/Morpho/MorphoAaveV3DebtTokenAdaptor.sol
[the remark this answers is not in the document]
0xEinCodes: Yep, we are good!

src/modules/adaptors/Morpho/MorphoAaveV3ATokenP2PAdaptor.sol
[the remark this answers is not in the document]
crispymangoes: Technically, balanceOf tells the cellar how much money it has there, withdrawableFrom tells the cellar how much money it CAN withdraw from the position. It is entirely possible for the 2 values to be the same.

0xEinCodes:
We've discussed any lingering comments offline. In particular those that were brought up by the morpho team (Lens contract). Nice work!

0xEinCodes (approved):

PR #121, opened 2023-06-06
Tweak logic so that it uses the rate provider instead of function sel…

crispymangoes:
…ectors on constituent

### DIFF
--- a/src/modules/price-router/Extensions/Balancer/BalancerStablePoolExtension.sol
+++ b/src/modules/price-router/Extensions/Balancer/BalancerStablePoolExtension.sol
@@ -2,6 +2,7 @@
 pragma solidity 0.8.16;
 
 import { IBalancerPool } from "src/interfaces/external/IBalancerPool.sol";
+import { IRateProvider } from "src/interfaces/external/IRateProvider.sol";
 import { BalancerPoolExtension, PriceRouter, ERC20, Math, IVault, IERC20 } from "./BalancerPoolExtension.sol";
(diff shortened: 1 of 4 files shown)


src/modules/price-router/Extensions/Balancer/BalancerStablePoolExtension.sol
```diff
@@ -116,13 +113,8 @@ contract BalancerStablePoolExtension is BalancerPoolExtension {
...
             uint256 price = priceRouter.getPriceInUSD(stor.underlyingOrConstituent[i]);
-            if (stor.functionSelectorToGetRate[i] != bytes4(0)) {
-                address constituent = address(stor.underlyingOrConstituent[i]);
-                bytes memory result = constituent.functionStaticCall(
-                    abi.encodeWithSelector(stor.functionSelectorToGetRate[i])
-                );
-                if (result.length == 0) revert BalancerStablePoolExtension__RateProviderCallFailed();
-                uint256 rate = abi.decode(result, (uint256));
+            if (stor.rateProviders[i] != address(0)) {
+                uint256 rate = IRateProvider(stor.rateProviders[i]).getRate();
```
0xEinCodes, line 117: We discussed if it is worthwhile to check that the `getRate()` return rate is within an acceptable range. This seems like a highly unlikely scenario that is out of scope for this PR because it kind of applies to general pricing risks, but we will open an issue.

One concept to kickoff this discussion in an issue: Pausing any cellars assuming the multisig can, then work with Strategists to rebalance out of it somehow in a way to mitigate losses.

PR #123, opened 2023-06-15
Feat/morpho is borrowing any

crispymangoes:
This PR fixes the Morpho Aave V2 debt token adaptor logic so that calling the strategist repay function will always repay the full amount of debt, where as before it would sometimes leave 1 wei of debt behind. 

In addition to the above it also removes the `isLiquid` configuration bool from the Morpho Aave V2 aToken adaptor, and instead reads the cellars `userMarkets` to determine if the cellar has any active loans. If the cellar has no active loans, the aToken position is liquid, otherwise the aToken position is illiquid.

### DIFF
--- a/src/modules/adaptors/Morpho/MorphoAaveV2ATokenAdaptor.sol
+++ b/src/modules/adaptors/Morpho/MorphoAaveV2ATokenAdaptor.sol
@@ -3,6 +3,7 @@ pragma solidity 0.8.16;
 
 import { BaseAdaptor, ERC20, SafeTransferLib, Math } from "src/modules/adaptors/BaseAdaptor.sol";
 import { IMorphoV2 } from "src/interfaces/external/Morpho/IMorphoV2.sol";
+import { IMorphoLensV2 } from "src/interfaces/external/Morpho/IMorphoLensV2.sol";
 import { MorphoRewardHandler } from "src/modules/adaptors/Morpho/MorphoRewardHandler.sol";
 import { IAaveToken } from "src/interfaces/external/IAaveToken.sol";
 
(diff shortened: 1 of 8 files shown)


0xEinCodes (approved):
Looks good to me, changes have been discussed with Auditor as well. The only thing I might ask for is to add a couple of lines in the main PR message here of what the intent of this PR is for future reference.

src/modules/adaptors/Morpho/MorphoAaveV2ATokenAdaptor.sol
```diff
@@ -97,12 +97,10 @@ contract MorphoAaveV2ATokenAdaptor is BaseAdaptor, MorphoRewardHandler {
     /**
      * @notice Uses configuration data to determine if the position is liquid or not.
      */
```
kingofclubstroyDev, line 119: This natSpec comment is not longer accurate

src/modules/adaptors/Morpho/MorphoAaveV2ATokenAdaptor.sol
```diff
@@ -97,12 +117,10 @@ contract MorphoAaveV2ATokenAdaptor is BaseAdaptor, MorphoRewardHandler {
...
-    function withdrawableFrom(
-        bytes memory adaptorData,
-        bytes memory configData
-    ) public view override returns (uint256) {
-        bool isLiquid = abi.decode(configData, (bool));
-        if (!isLiquid) return 0;
+    function withdrawableFrom(bytes memory adaptorData, bytes memory) public view override returns (uint256) {
```
0xEinCodes, line 120: @kingofclubstroyDev comment is right. Also please add params descriptions.

0xEinCodes (changes requested):
Looks good to me other than one change request on nat spec. You can address that with the other fixes from the audit. :)

PR #127, opened 2023-06-19
Update Cellar Adaptor to support illiquid and liquid positions.

crispymangoes:
When the `CellarAdaptor` was created it was mainly intended to be used by Cellars whose `asset` was the same asset used in the Cellar Position they were depositing into, or at the very least the two assets would be correlated. IE A Cellar denominated in FRAX could add the Real Yield USD Cellar as a position in it, so that it could move funds into Real Yield USD. So now a user withdrawing from the FRAX cellar could end up getting FRAX, USDC, USDT, or DAI from their withdraw. All these assets are more or less the same so users wouldn't really have an issue with this. 

However several strategies have emerged that no longer follow this correlated asset trend. Namely the Real Yield Gov Cellars. These Cellars `asset` is a Gov token such as LINK, but they can deposit funds into Real Yield ETH which is denominated in WETH. They do this by putting up LINK as collateral on Aave, borrowing WETH, then depositing it in Real Yield ETH. The problem with this is now a user withdraw can give them LINK, WETH, RETH, CBETH, or STETH. LINK is not directly correlated with any of these assets, leading to a user experience where a user must sell potentially 4 different tokens in order to get LINK.

To fix this, this PR adds an `isLiquid` bool to `CellarAdaptor.sol`s configuration data. This way the strategist can decide whether they want to restrict user withdraws from a Cellar position or not. 

The new logic is as follows.
`CellarAdaptor:withdraw()` Checks if `isLiquid` is false, if so it reverts.
`CellarAdaptor:withdrawableFrom()` Checks if `isLiquid` is true if so returns the positions `maxWithdraw` amount, otherwise it returns 0.


src/modules/adaptors/Sommelier/CellarAdaptor.sol
```diff
@@ -60,9 +67,18 @@ contract CellarAdaptor is BaseAdaptor {
      * @param assets the amount of assets to withdraw from the Cellar position
      * @param receiver address to send assets to'
      * @param adaptorData data needed to withdraw from the Cellar position
-     * @dev configurationData is NOT used
+     * @param configurationData abi encoded bool inidicating whether the position is liquid or not
```
kingofclubstroyDev, line 70: Spelling

crispymangoes: Done!

src/modules/adaptors/Sommelier/CellarAdaptor.sol
```diff
@@ -16,7 +16,14 @@ contract CellarAdaptor is BaseAdaptor {
     // Where:
     // `cellar` is the underling Cellar this adaptor is working with
     //================= Configuration Data Specification =================
-    // NOT USED
+    // configurationData = abi.encode(bool isLiquid)
+    // Where:
+    // `isLiquid` dictates whether the position is liquid or not
+    // If true:
+    //      position can support use withdraws
```
0xEinCodes, line 23: nit: "user" not "use"

src/modules/adaptors/Sommelier/CellarAdaptor.sol
```diff
@@ -16,7 +16,14 @@ contract CellarAdaptor is BaseAdaptor {
...
+    // else:
+    //      position can not support user withdraws
```
0xEinCodes, line 25: nit: "user" not "use"

0xEinCodes (approved):
Looks good, small spelling mistake in comments, otherwise good to go!

PR #133, opened 2023-07-28
Add stEth Extension, and Legacy Cellar Adaptor

crispymangoes:
The goal of the stEth Extension is to price stEth in a much less volatile way, without relying on a risky 1:1 peg between stEth:WETH.

The goal of the Legacy Cellar Adaptor is to allow Cellars to utilize ERC4626 Share Price Oracles when pricing Cellar positions, so that taking positions in other Cellars is much less gas intensive.

TODO
- [ ] Add a Balancer test where Flash Loan is actually used to do something.
- [ ] Address all compiler errors.


0xEinCodes:
As per async conversation with @crispymangoes, the remaining TODOs will be handled within other PRs after this is merged.

PR #136, opened 2023-08-07
Create DSR adaptor and add functionality tests

crispymangoes:
This PR creates the DAI Savings Rate Adaptor(`DSRAdaptor.sol`).

This adaptor allows Cellars to add and remove funds from the DSR.


src/modules/adaptors/Maker/DSRAdaptor.sol
```diff
@@ -0,0 +1,133 @@
...
+        return pieOf.mulDivDown(pot.chi(), 1e27);
+    }
+
+    /**
+     * @notice Returns the balance of DAI in the DSR.
+     * @dev Does not account for pending interest.
+     */
+    function balanceOf(bytes memory) public view override returns (uint256) {
+        uint256 pieOf = dsrManager.pieOf(msg.sender);
+        return pieOf.mulDivDown(pot.chi(), 1e27);
```
0xEinCodes, line 88: Is this okay to not account for pending interest? `chi` is updated upon `joins` and `exits` via [pot.drip](https://etherscan.deth.net/address/0x373238337Bfe1146fb49989fc222523f83081dDb) iirc.

Or are we not accounting for interest / earnings to be conservative?

test/testAdaptors/DSRAdaptor.t.sol
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    function testWithdraw(uint256 assets) external {
+        assets = bound(assets, 0.1e18, 1_000_000_000e18);
+        deal(address(DAI), address(this), assets);
+        cellar.deposit(assets, address(this));
+
+        uint256 maxRedeem = cellar.maxRedeem(address(this));
+
+        cellar.redeem(maxRedeem, address(this), address(this));
```
0xEinCodes, line 89: Add an `assert` to make sure that the cellar is getting back the right amount of assets.

test/testAdaptors/DSRAdaptor.t.sol
```diff
@@ -0,0 +1,215 @@
...
+
+        console.log("TA", cellar.totalAssets());
+
+        uint256 assetsBefore = cellar.totalAssets();
+
+        vm.warp(block.timestamp + 1 days);
+        mockDaiUsd.setMockUpdatedAt(block.timestamp);
+
+        assertEq(
+            cellar.totalAssets(),
```
0xEinCodes, line 105: Ah, I see, `balanceOf()` shouldn't have interest accounted for since it is a view function, not mutative.

test/testAdaptors/DSRAdaptor.t.sol
```diff
@@ -0,0 +1,215 @@
...
+            "Assets should not have increased because nothing has interacted with dsr."
+        );
+
+        uint256 maxRedeem = cellar.maxRedeem(address(this));
+        cellar.redeem(maxRedeem, address(this), address(this));
+
+        assertApproxEqAbs(DAI.balanceOf(address(this)), assets, 3, "Should have sent DAI to the user.");
+
+        uint256 bal = manager.daiBalance(address(cellar));
+        assertGt(bal, 0, "Balance should have left pending yield in DSR.");
```
0xEinCodes, line 140: Do we not want to pull accumulated yield as well from the DSR when carrying out a `redeem` w/ `maxRedeem` amounts?

0xEinCodes: We talked about this a bit in the meeting, but please elaborate. In general, it looks like the `dsrManager` exit call updates `chi` and thus the total exited amount has interest included. So wouldn't the `cellar` have the extra interest?

test/testAdaptors/DSRAdaptor.t.sol
```diff
@@ -0,0 +1,215 @@
...
+
+        // Withdraw remaining  assets.
+        {
+            bytes[] memory adaptorCalls = new bytes[](1);
+            adaptorCalls[0] = _createBytesDataToExitDsr(type(uint256).max);
+            data[0] = Cellar.AdaptorCall({ adaptor: address(dsrAdaptor), callData: adaptorCalls });
+        }
+        cellar.callOnAdaptor(data);
+
+        assertGt(
```
0xEinCodes, line 209: This is greater because we have accumulated interest right?

0xEinCodes (commented):
Overall looks good! It's a smaller adaptor, and pretty straight forward functionality. Please just respond regarding the interest accrual aspect. A relevant possible change request would be to add a function that updates the interest (`chi`) rate for the `DSR`. So it would be something like `callPotDrip()` or something.

src/modules/adaptors/Maker/DSRAdaptor.sol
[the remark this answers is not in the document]
crispymangoes: Yeah it is okay to not account for pending interest.

It really only becomes an issue if the pot is not interacted with for a long time, but strategists can call exit(1) and update it, or if a drip method is implemented that would work too.

Adaptors not accounting for pending interest has worked just fine in FraxLend V1 positions, and cToken positions in production, so I don't see any major issues with this.

test/testAdaptors/DSRAdaptor.t.sol
crispymangoes: No `exit` will only pull out the amount of assets you specify, so the pending yield is left in the DSR, which is fine, the strategist can handle that however they want to during a rebalance.

0xEinCodes: Roger that, looks good then! Thanks for the clarification.

src/modules/adaptors/Maker/DSRAdaptor.sol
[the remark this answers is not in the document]
0xEinCodes: Sounds good, thanks for the clarification! It looks like you went with the strategist calling drip() prior which works.

0xEinCodes (approved):
Looks good to me. Nice work!

PR #137, opened 2023-08-08
Remove `positionCount`, from Registry, and add missing Distrust Posi…

crispymangoes:
…tion Event


0xEinCodes:
Looks good. nit - spelling mistake in your commit.

PR #145, opened 2023-10-02
Dev/macro audit 9

crispymangoes:
This PR includes changes from Macro Audits 9, 10, and 11. Also multiple integration tests, and a few deploy script modifications for common changes.


src/interfaces/IRYUSDRegistry.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.21;
+
+/**
+ * @title IRYUSDRegistry
+ * @author
```
0xEinCodes, line 6: nit: missing author

0xEinCodes: probably just remove author and natspec from this like other interfaces.

PR #149, opened 2023-10-16
Feat/erc4626 share price oracle automation forwarder

crispymangoes:


0xEinCodes (commented):
All in all looks good to me. I noticed you commented out some test code in the `LegacyCellarAdaptorForRye`, may you expand on that? Also I have a couple clarification questions that I'm not too worried about:

1. Interfaces are custom? I may have just missed this at a glance but it seems `getForwarder` is not an interface function within `IRegistry.sol` from Chainlink for example.
2. `Forwarder` == reverse proxy. Essentially it acts as a single address that any node can report to. Is this an alright interpretation? This would make sense that it has to be the one performing `upkeep` now.

crispymangoes:
Yeah so `LegacyCellarAdaptorForRye` is a bit weird. The original test relies on blockchain state from a couple months ago. But the new Chainlink Automation V2 contracts are not deployed at said block. So in order for the new share price oracle contracts to work we need a much more recent blockchain state, but this new blockchain state already has the RYBTC, RYE integration completed, so the test state is no longer valid.

Because of this the test is commented out. In the future we might completely remove this test, but for now it has been left commented out, so that it can be referenced.

PR #151, opened 2023-10-22
Add test showing sFRAX integration is supported using existing Cellar Adaptor

crispymangoes:
As title says, this PR adds a test where we use the existing Cellar Adaptor to interact with sFRAX. Specifically checking that strategists can move funds in/out of sFRAX, and that users can deposit/withdraw from sFRAX.


0xEinCodes (changes requested):
Looks good to me. One thing I am wondering about is what happens if the cellar withdraws its entire amount from the sFRAX position prior to the weekly Wednesday deposit of Frax from the FRAX protocol as per their [docs](https://docs.frax.finance/frax-v3/sfrax#:~:text=Every%20Wednesday%20at%2011%3A59%3A59%20UTC%20the%20Frax%20Protocol%20adds%20newly%20minted%20FRAX%20stablecoins%20into%20the%20sFRAX%20vault.). Once this is investigated, or if you can just comment on this, then I think we can go ahead and merge this.

crispymangoes:
So the share for the sFRAX vault increases in value every single block. My best guess is that when they do that weekly deposit, they distribute the rewards out over the course of a week(or some preset time), so that people can not game the system and deposit right before new rewards are added. So to answer your question I dont really think anything will happen, the Cellar will continue earning yield until it redeems all of its shares.

0xEinCodes:
> So the share for the sFRAX vault increases in value every single block. My best guess is that when they do that weekly deposit, they distribute the rewards out over the course of a week(or some preset time), so that people can not game the system and deposit right before new rewards are added. So to answer your question I dont really think anything will happen, the Cellar will continue earning yield until it redeems all of its shares.

We discussed async and I think we can proceed. We'll simply notify the strategists using this adaptor that it is best to wait for the payout block (every Wednesday atm). Thanks!

0xEinCodes (approved):

PR #152, opened 2023-10-27
Fix/macro audit 12

crispymangoes:
Fixes M1, M2, L1, L2


0xEinCodes:
All looks good to me. The external auditor has looked it over from my understanding which adds more review and due diligence. Nice work!

0xEinCodes (approved):

PR #154, opened 2023-11-03
Design, Develop, and Test Morpho Blue Adaptors

0xEinCodes:
## Core changes:

This PR encompasses the development and unit testing of the MorphoBlue supply, collateral, and debt token adaptors.

There are three different phases/scopes. They are as follows:
- Scope 1: `MorphoBlueSupplyAdaptor.sol` - Allows supply positions to Morpho Blue Lending Markets.
- Scope 2: `MorphoBlueCollateralAdaptor.sol`  - Allows collateral provision positions to Morpho Blue Lending Markets.
- Scope 3: `MorphoDebtAdaptor.sol` - Allows debt positions to be taken from  Morpho Blue Lending Markets. 

Extra: HealthFactor Logic smart contracts that likely will be used by the `CollateralMorphoBlueAdaptor.sol` and `DebtMorphoBlueAdaptor.sol`

### High-Level Key Points:

1. Morpho Blue is a lending protocol comprised of isolated market pairs. These are all permissionless, and use singleton design; within the Morpho Blue repo, [`Morpho.sol`](https://github.com/morpho-org/morpho-blue/blob/main/src/Morpho.sol) is the singleton contract that keeps track of all markets created, as well as associated accounting.
2. Periphery libraries are created and tested by Morpho team for external protocols, like Sommelier, to have getters available for accounting and easier integration. For now, these adaptor designs do not use the specific periphery library [`MorphoBalancesLib.sol`](https://github.com/morpho-org/morpho-blue/blob/main/src/libraries/periphery/MorphoBalancesLib.sol) to be 100% sure we are pulling real values from storage off of prod contract `Morpho.sol` for Morpho Blue markets. We may switch to using periphery libraries later, or within this audit after change requests, whichever comes first.
3. Tests have been written with a mix of mock markets and IRMs since Morpho Blue is deploying about a week before the scheduled audit for these adaptors. Due to these adaptors being released basically at the launch of Morpho Blue, additional review from the Morpho team is requested.

### Discussion Points

1. Currently there is underflow/overflow errors occurring within `MorphoSupplyAdaptor.t.sol` in `testWithdraw()`. See [here](https://github.com/PeggyJV/cellar-contracts/blob/0b7a5f6074504b3563868c79c6a424cdaa691a7d/test/testAdaptors/MorphoBlue/MorphoBlueSupplyAdaptor.t.sol#L288).

### References:
1. Morpho Blue WhitePaper: https://github.com/morpho-org/morpho-blue/blob/main/morpho-blue-whitepaper.pdf 
2. Morpho Blue Codebase: https://github.com/morpho-org/morpho-blue 

---
<details markdown='1'><summary> ✅ TODOs:
 </summary>

### Scope 1: `MorphoBlueSupplyAdaptor.sol`
- [x] Stub out code for supplying loanToken to MorphoBlue Lending Markets
- [x] Add strategist functions to supply loanToken
- [x] Add base functions
- [x] Debug compilation errors and reformat
- [x] Write unit tests
- [x] reformat and remove unneeded comments and TODOs
- [x] Full PR Review from Crispy. 
- [x] Change Requests
- [ ] Finalize

### Scope 2: `MorphoBlueCollateralAdaptor.sol`
- [x] Stub out code
- [x] Add strategist functions to deposit and remove collateral
- [x] Add base functions
- [x] Write unit tests
- [x] reformat and remove unneeded comments and TODOs
- [x] Full PR Review from Crispy. 
- [x] Change Requests
- [ ] Finalize

### Scope 3: `MorphoDebtAdaptor.sol` TODOs:

- [x] Stub out code w/ referenced docs on how to take debt while keeping track of health factor
- [x] Add strategist functions (borrow, repay, accrueInterest, etc.).
- [x] Add base functions
- [x] Write unit tests
- [x] reformat and remove unneeded comments and TODOs
- [x] Full PR Review from Crispy. 
- [x] Change Requests
- [ ] Finalize

### External TODOs:
- [x] Send PR off to Morpho team for external review.
- [ ] Make any changes as needed.

</details>

<!--- example - Design decisions made in PR>
<!--- example - Remember running `forge fmt` before committing the PR>
<!--- example - Reference for how to run relevant tests i.e. forge test --match-contract X --fork-url $RPC_URL -vvv>


0xEinCodes:
This is a very rough draft PR. It's more made to put an ear mark on the research done to better understand Morpho Blue and how the architecture of the adaptor would look. As one can see, the draft PR has tons of aspects still from Fraxlend adaptor implementation (and even comments).

I will come back to this and clean things up so it only references Morpho Blue. Just a bit late where I'm at 😴.

0xEinCodes:
Updates done this week:

- Supply adaptor rough implementation completed.
- Health Factor logic calculations
   - Gone through Morpho Blue codebase, assessed how health factor calculations would work using similar mathematical methods / sequence as they do within `_isHealthy()`.
- Implement calls to `getHealthFactor()` within helper contract throughout supply, and collateral adaptors. Currently finishing up things in debt adaptor.


TODO next week:
- [ ] Debug compilation errors
- [ ] Reformat (remove mentions of Frax from old Fraxlend adaptors)
- [ ] Write tests for review in the new year prior to audit on January 15th, 2024

NOTE:
- `MorphoBlueHealthFactorLogic.sol` doesn’t use the exposed getters from the MorphoBlue periphery libraries.

0xEinCodes:
Today's Updates:
- Write Unit Tests: Create setup for collateral and debt tests, write collateral adaptor tests, start writing debt adaptor tests

NOTES / Questions to Discuss:
- Do we want to have `accrueInterest()` as a Strategist function within all adaptors for Morpho Blue, or just don't have it in the `MorphoBlueDebtAdaptor.sol` contract since in order to have debt, the cellar would need to have a collateral position already and that adaptor would have `accrueInterst()` as a strategist function.
- Commented out code within tests are from Fraxlend since those adaptors have similar architectural design.

TODO:
- [ ] For test deployment of MorphoBlue && interacting with appropriate market Ids, reference how Morpho Blue went about their tests.
- [ ] Write rest of tests tomorrow ( `morphoBlueDebtAdaptor`, `morphoBlueSupplyAdaptor`, and combination tests).

0xEinCodes:
Friday DONE:
- [x] Write `morphoBlueSupplyAdaptor` tests
- [ ] Start writing the setup we will use (for morphoBlue deployment and marketIds for our tests).

Tuesday (Jan 2) TODO:
- [ ] Finish writing the setup we will use (for morphoBlue deployment and marketIds for our tests).
- [ ] Debug to get test suite running
- [ ] Assess gas efficiencies btw using periphery libraries and not

Wed/Thurs/Friday (Jan 3-5) TODO:
- [ ] PR Review from Crispy
- [ ] PR Review CRs

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueSupplyAdaptor.sol
```diff
@@ -0,0 +1,254 @@
...
+     * @notice Attempted to interact with a Morpho Blue Lending Market that the Cellar is not using.
+     */
+    error MorphoBlueSupplyAdaptor__MarketPositionsMustBeTracked(Id id);
+
+    /**
+     * @notice This bool determines how this adaptor accounts for interest.
+     *         True: Account for pending interest to be paid when calling `balanceOf` or `withdrawableFrom`.
+     *         False: Do not account for pending interest to be paid when calling `balanceOf` or `withdrawableFrom`.
+     */
+    bool public immutable ACCOUNT_FOR_INTEREST;
```
crispymangoes, line 44: This is not used

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueSupplyAdaptor.sol
```diff
@@ -0,0 +1,254 @@
...
+
+    /**
+     * @notice Allows strategists to withdraw underlying asset plus interest.
+     */
+    function withdrawFromMorphoBlue(Id _id, uint256 _assets) public {
+        // Run external receiver check.
+        _externalReceiverCheck(address(this));
+        _validateMBMarket(_id);
+        MarketParams memory market = morphoBlue.idToMarketParams(_id);
+        _accrueInterest(market); // TODO - if we end up using periphery library for _balanceOf() then we may not need to kick `accrueInterest()`. We sacrifice losing some dust / noise though I think. Need to test this.
```
crispymangoes, line 181: I think it is best to leave the accrueInterest call in here. But I would move it down into the if statement. Cuz if the strategist is not trying to pull everything from it, then we dont need to kick it before hand cuz we know exactly how much the strategist wants to pull

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueSupplyAdaptor.sol
```diff
@@ -0,0 +1,254 @@
...
+        MarketParams memory market = morphoBlue.idToMarketParams(id);
+        _withdraw(market, assets, address(this));
+    }
+
+    /**
+     * @notice Returns the amount of loanToken that can be withdrawn.
+     * @dev Compares loanToken supplied to loanToken borrowed to check for liquidity.
+     *      - If loanToken balance is greater than liquidity available, it returns the amount available.
+     * @param adaptorData encoded bytes32 MB id that represents the MB market for this position.
+     * @return withdrawableSupply liquid amount of `loanToken` cellar has lent to specified MB market.
```
crispymangoes, line 109: So if the simmed values use a lot of gas to generate (50-100k) then I think we just use the less correct getters from Morpho.sol. But if they have a low gas overhead, then the simmed values are fine

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueCollateralAdaptor.sol
```diff
@@ -0,0 +1,230 @@
...
+
+/**
+ * @title Morpho Blue Collateral Adaptor
+ * @notice Allows addition and removal of collateralAssets to Morpho Blue pairs for a Cellar.
+ * @dev This adaptor is specifically for Morpho Blue Primitive contracts.
+ *      To interact with a different version or custom market, a new
+ *      adaptor will inherit from this adaptor
+ *      and override the interface helper functions. MB refers to Morpho
+ *      Blue
+ * @author crispymangoes, 0xEinCodes
```
crispymangoes, line 16: If you want you can flip the names, or even remove my name entirely, since this is all your work, and I am just reviewing it :)

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueCollateralAdaptor.sol
```diff
@@ -0,0 +1,230 @@
...
+     * @param assets the amount of assets to provide as collateral on Morpho Blue
+     * @param adaptorData adaptor data containing the abi encoded Id for specific Morpho Blue Lending Market
+     * @dev configurationData is NOT used
+     */
+    function deposit(uint256 assets, bytes memory adaptorData, bytes memory) public override {
+        // Deposit assets to Morpho Blue.
+        Id id = abi.decode(adaptorData, (Id));
+        _validateMBMarket(id);
+
+        MarketParams memory market = morphoBlue.idToMarketParams(id);
```
crispymangoes, line 76: How much state does this read? I wonder if it would be better to just have the adaptor data store the ID and the loan token

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueCollateralAdaptor.sol
```diff
@@ -0,0 +1,230 @@
...
+    /**
+     * @notice Allows strategists to remove collateral from the respective cellar position on specified MB Market.
+     * @param _id encoded bytes32 MB id that represents the MB market for this position.
+     * @param _collateralAmount The amount of collateral to remove from specified MB market position.
+     */
+    function removeCollateral(Id _id, uint256 _collateralAmount) public {
+        _validateMBMarket(_id);
+        MarketParams memory market = morphoBlue.idToMarketParams(_id);
+
+        _accrueInterest(market);
```
crispymangoes, line 163: Move inside if statement, see supply adaptor comment

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueCollateralAdaptor.sol
```diff
@@ -0,0 +1,230 @@
...
+        if (_collateralAmount == type(uint256).max) {
+            _collateralAmount = _userCollateralBalance(_id, address(this));
+        }
+
+        // remove collateral
+        _removeCollateral(market, _collateralAmount);
+
+        // TODO - might want to check the market to see if it even has a LLTV. If it doesn't, I guess no liquidations can occur?
```
crispymangoes, line 171: So if it doesn't have an LLTV, then we can not take on any borrows. So I think that it is okay to remove this check, because we will only add a 0 LLTV market to the registry, if we want to use them. But for now I dont see any use of them for us.

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueDebtAdaptor.sol
```diff
@@ -0,0 +1,244 @@
...
+     * @notice Attempted repayment when no debt position in Morpho Blue Lending Market for cellar
+     */
+    error MorphoBlueDebtAdaptor__CannotRepayNoDebt(Id id);
+
+    /**
+     * @notice This bool determines how this adaptor accounts for interest.
+     *         True: Account for pending interest to be paid when calling `balanceOf` or `withdrawableFrom`.
+     *         False: Do not account for pending interest to be paid when calling `balanceOf` or `withdrawableFrom`.
+     */
+    bool public immutable ACCOUNT_FOR_INTEREST;
```
crispymangoes, line 53: This is unused

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueDebtAdaptor.sol
```diff
@@ -0,0 +1,244 @@
...
+        uint256 totalBorrowShares = morphoBlue.market(_id).totalBorrowShares;
+        uint256 sharesToRepay = debtAmountToRepay.toSharesUp(totalBorrowAssets, totalBorrowShares); // get the total assets and total borrow shares of the market
+
+        // MorphoBlue reverts w/ underflow/overflow error if trying to repay with more than we have. That said, we will manage when strategists try to pass in type(uint256).max
+        uint256 sharesAccToMorphoBlue = morphoBlue.borrowShares(_id, address(this));
+        if (sharesAccToMorphoBlue < sharesToRepay) {
+            sharesToRepay = sharesAccToMorphoBlue;
+        }
+        
+        tokenToRepay.safeApprove(address(morphoBlue), type(uint256).max);
```
crispymangoes, line 184: Wouldn't the amount be `debtAmountToRepay` instead of a `type(uint256).max`? If there is weird rounding errors, then its fine to use the uint256 max logic since we revoke approval afterwards, but if we do know the exact amount of assets morpho needs to repay the debt, then we should use that.

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueDebtAdaptor.sol
```diff
@@ -0,0 +1,244 @@
...
+    /**
+     * @notice Allows a strategist to call `accrueInterest()` on a MB Market cellar is using.
+     * @dev A strategist might want to do this if a MB market has not been interacted with
+     *      in a while, and the strategist does not plan on interacting with it during a
+     *      rebalance.
+     * @dev Calling this can increase the share price during the rebalance,
+     *      so a strategist should consider moving some assets into reserves.
+     * @param _id encoded bytes32 MB id that represents the MB market for this position.
+     */
+    function accrueInterest(Id _id) public {
```
crispymangoes, line 198: Can this logic be moved into a helper file?

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueHealthFactorLogic.sol
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * @notice The ```_getHealthFactor``` function returns the current health factor of a respective position given an exchange rate
+     * @param _id The specified Morpho Blue market Id
+     * @param _market The specified Morpho Blue market
+     * @return currentHF The health factor of the position atm
+     */
+    function _getHealthFactor(Id _id, MarketParams memory _market) internal view virtual returns (uint256 currentHF) {
+        uint256 borrowAmount = _userBorrowBalance(_id, address(this));
+        console.log("borrowAmount: %s", borrowAmount);
+        if (borrowAmount == 0) return 1.05e18; // TODO - decide what to return in these scenarios.
```
crispymangoes, line 52: I would return `type(uint256).max`. That is the convention Aave uses.

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueHealthFactorLogic.sol
```diff
@@ -0,0 +1,122 @@
...
+        morphoBlue = IMorpho(_morphoBlue);
+    }
+
+    /**
+     * @notice The ```_getHealthFactor``` function returns the current health factor of a respective position given an exchange rate
+     * @param _id The specified Morpho Blue market Id
+     * @param _market The specified Morpho Blue market
+     * @return currentHF The health factor of the position atm
+     */
+    function _getHealthFactor(Id _id, MarketParams memory _market) internal view virtual returns (uint256 currentHF) {
```
crispymangoes, line 49: This function does not kick the market because the interest should already be accrued from some strategist action before right? Like taking on a new borrow, or repaying debt, or removing collateral?

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueHealthFactorLogic.sol
```diff
@@ -0,0 +1,122 @@
...
+                market.totalBorrowShares
+            )
+        );
+    }
+
+    /**
+     * @notice Caller calls `accrueInterest` on specified MB market
+     * @param _market The specified MB market
+     */
+    function _accrueInterest(MarketParams memory _market) internal virtual {
```
crispymangoes, line 119: This accrueInterest is not validating market input.

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueSupplyAdaptor.sol
```diff
@@ -0,0 +1,254 @@
...
+     */
+    function withdraw(uint256 assets, address receiver, bytes memory adaptorData, bytes memory) public override {
+        // Run external receiver check.
+        _externalReceiverCheck(receiver);
+
+        // Withdraw assets from Morpho Blue.
+        Id id = abi.decode(adaptorData, (Id));
+        _validateMBMarket(id);
+        MarketParams memory market = morphoBlue.idToMarketParams(id);
+        _withdraw(market, assets, address(this));
```
crispymangoes, line 101: I think `address(this)` should be `receiver`, so that the assets are transferred to the user, as opposed to the Cellar.

test/testAdaptors/MorphoBlue/MorphoBlueSupplyAdaptor.t.sol
```diff
@@ -0,0 +1,689 @@
...
+
+        assertEq(assetsWithdrawable, liquidLoanToken2, "Should be able to withdraw liquid loanToken.");
+
+        // Have user withdraw the loanToken.
+        deal(address(USDC), address(this), 0);
+        cellar.withdraw(liquidLoanToken2, address(this), address(this));
+        assertEq(USDC.balanceOf(address(cellar)), liquidLoanToken2, "User should have received liquid loanToken."); // TODO - Crispy, this should be address(this), but it seems to just be unwinding from morphoBlue and not safetransferring to the test address as it should be doing from morphoBlue (see their codebase for withdraw()).
+    }
+
+    // TODO - Review with Crispy as the pricing is stale although I tried troubleshooting with setMockUpdatedAt.As well, other tests with whale didn't throw stale pricing reverts
```
crispymangoes, line 603: You are updating the timestamp of the mock usdc feed, but I think you also need to update the timestamp of the mock dai feed, cuz the error message is saying DAI pricing is stale, not USDC

crispymangoes (changes requested):

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueSupplyAdaptor.sol
0xEinCodes: Removed!

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueSupplyAdaptor.sol
0xEinCodes: I'll leave this comment open, but am deleting comments within the codebase about the periphery libraries. We can bring it back into discussion if it is found worthwhile as discussed.

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueSupplyAdaptor.sol
0xEinCodes: Sounds good. I've moved it into the if statement logic, and since the morphoBlue codebase should revert when trying to withdraw more than one has supplied, this should work out fine. I ran into an unexpected overflow/underflow though when adding more checks to `testWithdraw()` that I think might be worthwhile to check with Morpho team this week.

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueCollateralAdaptor.sol
0xEinCodes: Haha it doesn't really matter to me, but I'll flip them in case people have questions and can direct them to me first.

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueCollateralAdaptor.sol
0xEinCodes: I'll have to look into this, but that might just be an easier thing to do and could reduce gas.

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueCollateralAdaptor.sol
0xEinCodes: Hmm, I actually think we don't need `_accrueInterest()` in this function call. Collateral is not affected by interest accrual, just the health factor calculation. Removing collateral also calls `accrueInterest()` in morphoBlue contracts so it's already done within the mutative part of the function call.

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueCollateralAdaptor.sol
0xEinCodes: sg2m!

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueDebtAdaptor.sol
0xEinCodes: Removed

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueDebtAdaptor.sol
0xEinCodes: Yeah it would be `debtAmountToRepay` if `sharesAccToMorphoBlue > sharesToRepay `. Either way though, approving what the cellar actually has is the most that can be done. I'll change it to `debtAmountToRepay`

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueHealthFactorLogic.sol
0xEinCodes: It validates it within the external `accrueInterst()` calls that then call this internal function. I looked at bringing over the `accrueInterest()` function over as a helper that is external but since it calls `validateMBMarket` for each respective adaptor, it seems that it could be simpler / work with separate external `accrueInterest()` calls within each adaptor.

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueDebtAdaptor.sol
0xEinCodes: See my comment [here](https://github.com/PeggyJV/cellar-contracts/pull/154/files#r1443878990).

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueHealthFactorLogic.sol
0xEinCodes: Also I do agree that putting `accrueInterest` as a helper would make sense overall (if the validate market aspect wasn't in the way). Although that would mean the `morphoBlueCollateralAdaptor` would also have the `accrueInterest()` function. This is fine, just a bit redundant, but I guess since a cellar would likely have a position with collateral in order to take out debt, then we're good. The aspect of zero LLTV markets might also bring up a different scenario, but that's TBD I guess.

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueHealthFactorLogic.sol
0xEinCodes: Correct, this should not kick the market as other functions that subsequently kick the market happen beforehand (borrow, repay, remove collateral). The mentioned actions would be the ones calling the health factor check, whereas things like adding collateral, or lending, or withdrawing loan tokens would not call this.

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueHealthFactorLogic.sol
0xEinCodes: Good call, I've made the change.

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueSupplyAdaptor.sol
0xEinCodes: Agreed. I've made the change!

test/testAdaptors/MorphoBlue/MorphoBlueSupplyAdaptor.t.sol
0xEinCodes: This has been updated as per our conversation outside of this PR. The test is working now :)

0xEinCodes:
@crispymangoes thanks for the review! I've reviewed your comments and responded to all of them. There is one test for checking `balanceOf()` within the supply adaptors, and then there is just reformatting. Otherwise this is ready to go if you are good with it after looking over my responses to your PR review comments. :)

src/interfaces/external/Morpho/MorphoBlue/libraries/ConstantsLib.sol
Rubilmax, line 1: `ConstantsLib` is a library stored under `src/interfaces`. Is it intended?

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueCollateralAdaptor.sol
Rubilmax: This call to `idToMarketParams` should cost 5 cold SLOADs, so 10k alone (the storage slots loaded here are not used by Blue anywhere). Ideally yes, the required data is encoded in `adaptorData`, so it costs much less (this is the design we've chosen for Blue)

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueCollateralAdaptor.sol
```diff
@@ -0,0 +1,224 @@
...
+     */
+    function addCollateral(Id _id, uint256 _collateralToDeposit) public {
+        _validateMBMarket(_id);
+        MarketParams memory market = morphoBlue.idToMarketParams(_id);
+        ERC20 collateralToken = ERC20(market.collateralToken);
+        uint256 amountToDeposit = _maxAvailable(collateralToken, _collateralToDeposit);
+        collateralToken.safeApprove(address(morphoBlue), amountToDeposit);
+        _addCollateral(market, amountToDeposit);
+        // Zero out approvals if necessary.
+        _revokeExternalApproval(collateralToken, address(morphoBlue));
```
Rubilmax, line 151: This logic is very similar to `deposit` aside from the incoming transfer from `msg.sender`. Would it be interesting to factorize this logic?

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueHelperLogic.sol
```diff
@@ -0,0 +1,111 @@
...
+    using SharesMathLib for uint256;
+    using MorphoLib for IMorpho;
+
+    /**
+     * @notice The Morpho Blue contract on current network.
+     */
+    IMorpho public immutable morphoBlue;
+
+    // Constant from Morpho Blue
+    uint256 constant ORACLE_PRICE_SCALE = 1e36;
```
Rubilmax, line 32: This constant is available in `ConstantsLib`

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueHelperLogic.sol
```diff
@@ -0,0 +1,111 @@
...
+    function _userSupplyBalance(Id _id, address _user) internal view returns (uint256) {
+        Market memory market = morphoBlue.market(_id);
+        return (morphoBlue.supplyShares(_id, _user).toAssetsDown(market.totalSupplyAssets, market.totalSupplyShares));
+    }
+
+    /**
+     * @dev helper function that returns actual collateral position amount for specified `_user` according to MB market accounting. This is alternative to using the MB periphery libraries that simulate accrued interest balances.
+     * @param _id identifier of a Morpho Blue market.
+     * @param _user address that this function will query Morpho Blue market for.
+     * NOTE: make sure to call `accrueInterest()` on respective market before calling these helpers.
```
Rubilmax, line 82: This is not required for collateral as collateral does not accrue interest

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueHelperLogic.sol
```diff
@@ -0,0 +1,111 @@
...
+
+    /**
+     * @dev helper function that returns actual supply position amount for specified `_user` according to MB market accounting. This is alternative to using the MB periphery libraries that simulate accrued interest balances.
+     * @param _id identifier of a Morpho Blue market.
+     * @param _user address that this function will query Morpho Blue market for.
+     * @return Actual supply amount for the `_user`
+     * NOTE: make sure to call `accrueInterest()` on respective market before calling these helpers.
+     */
+    function _userSupplyBalance(Id _id, address _user) internal view returns (uint256) {
+        Market memory market = morphoBlue.market(_id);
```
Rubilmax, line 74: Note you could save 2 SLOAD by using `extsloads` from Blue and `MorphoStorageLib`. But it most cases, these slots should be warm so it should not save much

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueHelperLogic.sol
```diff
@@ -0,0 +1,111 @@
...
+        uint256 collateralPrice = IOracle(_market.oracle).price(); // recall from IOracle.sol that the units will be 10 ** (36 - collateralUnits + borrowUnits) BUT collateralPrice is in units of borrow.
+
+        // get collateralAmount in borrowAmount for LTV calculations
+        uint256 collateralAmount = _userCollateralBalance(_id, address(this));
+        console.log(
+            "userCollateralBalance: %s, collateralPrice: %s, collateralAmount: %s",
+            _userCollateralBalance(_id, address(this)),
+            collateralPrice,
+            collateralAmount
+        );
```
Rubilmax, line 56: This should be dropped right?

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueHelperLogic.sol
```diff
@@ -0,0 +1,111 @@
...
+        uint256 collateralAmountInBorrowUnits = collateralAmount.mulDivDown(collateralPrice, ORACLE_PRICE_SCALE);
+
+        if (collateralAmountInBorrowUnits == 0) return 0;
+        uint256 LTV_PRECISION = 1e18;
+        uint256 currentPositionLTV = borrowAmount.mulDivUp(LTV_PRECISION, collateralAmountInBorrowUnits);
+        // convert LTVs to HF
+        currentHF = (_market.lltv).mulDivDown(1e18, currentPositionLTV);
```
Rubilmax, line 63:
```suggestion
        currentHF = _market.lltv.mulDivDown(collateralAmountInBorrowUnits, borrowAmount);
```

should be more precise and avoids an early return (complex logic)

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueCollateralAdaptor.sol
Jean-Grimal: It is also possible to have the adaptor data store the MarketParams instead. The call to `MarketParamsLib.id(marketParams)` is way cheaper than the call to `idToMarketParams(id)` as it doesn't read storage.

Jean-Grimal (changes requested):

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueSupplyAdaptor.sol
```diff
@@ -0,0 +1,238 @@
...
+     * @return withdrawableSupply liquid amount of `loanToken` cellar has lent to specified MB market.
+     */
+    function withdrawableFrom(
+        bytes memory adaptorData,
+        bytes memory
+    ) public view override returns (uint256 withdrawableSupply) {
+        Id id = abi.decode(adaptorData, (Id));
+        MarketParams memory market = morphoBlue.idToMarketParams(id);
+        (uint256 totalSupplyAssets, , uint256 totalBorrowAssets, ) = morphoBlue.expectedMarketBalances(market);
+        if (totalBorrowAssets >= totalSupplyAssets) return 0;
```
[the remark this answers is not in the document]
MerlinEgalite: not sure we can say that it will save gas given it's very unlikely to be in this specific situation

MerlinEgalite (commented):
Do you have the full trace about this issue? https://github.com/PeggyJV/cellar-contracts/blob/0b7a5f6074504b3563868c79c6a424cdaa691a7d/test/testAdaptors/MorphoBlue/MorphoBlueSupplyAdaptor.t.sol#L288

src/interfaces/external/Morpho/MorphoBlue/libraries/ConstantsLib.sol
0xEinCodes: Brought over libraries from Morpho repo as we started this integration effort. Removing it and unused libraries / interface files now.

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueCollateralAdaptor.sol
0xEinCodes: Good catch, thanks for this, yes factorizing is interesting. I've implemented some factorizing. 

NOTE: it can be further refactored if we use `marketParams` as an encoded adaptorData param. I will continue looking into it.

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueHelperLogic.sol
0xEinCodes: True, but I've removed the `ConstantsLib` now since we would only use the `ORACLE_PRICE_SCALE` constant from it.

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueHelperLogic.sol
0xEinCodes: Thanks! Removed the nat spec that does not apply.

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueHelperLogic.sol
0xEinCodes: Correct, I've removed it.

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueHelperLogic.sol
0xEinCodes: I think this works, I've added it to the changes. Thank you!

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueHelperLogic.sol
```diff
@@ -0,0 +1,111 @@
...
+        console.log(
+            "userCollateralBalance: %s, collateralPrice: %s, collateralAmount: %s",
+            _userCollateralBalance(_id, address(this)),
+            collateralPrice,
+            collateralAmount
+        );
+        uint256 collateralAmountInBorrowUnits = collateralAmount.mulDivDown(collateralPrice, ORACLE_PRICE_SCALE);
+
+        if (collateralAmountInBorrowUnits == 0) return 0;
+        uint256 LTV_PRECISION = 1e18;
```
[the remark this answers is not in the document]
0xEinCodes: This has been removed as per this new [implementation suggestion](https://github.com/PeggyJV/cellar-contracts/pull/154#discussion_r1445763062).

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueSupplyAdaptor.sol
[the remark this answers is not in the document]
0xEinCodes: I've made the changes nonetheless. It is fair that `totalBorrowAssets !> totalSupplyAssets`

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueSupplyAdaptor.sol
```diff
@@ -0,0 +1,238 @@
...
+     * @param _assets the amount of loanToken to withdraw from MB market
+     */
+    function withdrawFromMorphoBlue(Id _id, uint256 _assets) public {
+        // Run external receiver check.
+        _externalReceiverCheck(address(this));
+        _validateMBMarket(_id);
+        MarketParams memory market = morphoBlue.idToMarketParams(_id);
+        if (_assets == type(uint256).max) {
+            _accrueInterest(market);
+            _assets = _userSupplyBalance(_id, address(this));
```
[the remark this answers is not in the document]
0xEinCodes: Hmmm. The helper `_userSupplyBalance()` ultimately queries for the shares and converts them as seen below (it is in the `MorphoBlueHelperLogic.sol` file). I think this suffices unless there are more gas-efficient methods.

```
function _userSupplyBalance(Id _id, address _user) internal view returns (uint256) {
        Market memory market = morphoBlue.market(_id);
        return (morphoBlue.supplyShares(_id, _user).toAssetsDown(market.totalSupplyAssets, market.totalSupplyShares));
    }
```

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueDebtAdaptor.sol
```diff
@@ -0,0 +1,227 @@
...
+        ERC20 tokenToRepay = ERC20(market.loanToken);
+        uint256 debtAmountToRepay = _maxAvailable(tokenToRepay, _debtTokenRepayAmount);
+        // using Morpho sharesLibrary we can calculate the sharesToRepay from the debtAmount
+        uint256 totalBorrowAssets = morphoBlue.market(_id).totalBorrowAssets;
+        uint256 totalBorrowShares = morphoBlue.market(_id).totalBorrowShares;
+        uint256 sharesToRepay = debtAmountToRepay.toSharesUp(totalBorrowAssets, totalBorrowShares);
+        uint256 sharesAccToMorphoBlue = morphoBlue.borrowShares(_id, address(this));
+        if (sharesAccToMorphoBlue < sharesToRepay) {
+            sharesToRepay = sharesAccToMorphoBlue;
+        }
```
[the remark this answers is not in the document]
0xEinCodes: Thank you for this comment! I've made changes as per your suggestion. It makes sense to me.

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueHelperLogic.sol
```diff
@@ -0,0 +1,111 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.21;
+
+import { IMorpho, MarketParams, Market, Id } from "src/interfaces/external/Morpho/MorphoBlue/interfaces/IMorpho.sol";
+import { MathLib, WAD } from "src/interfaces/external/Morpho/MorphoBlue/libraries/MathLib.sol";
+import { SharesMathLib } from "src/interfaces/external/Morpho/MorphoBlue/libraries/SharesMathLib.sol";
+import { IOracle } from "src/interfaces/external/Morpho/MorphoBlue/interfaces/IOracle.sol";
+import { UtilsLib } from "src/interfaces/external/Morpho/MorphoBlue/libraries/UtilsLib.sol";
+import { MorphoLib } from "src/interfaces/external/Morpho/MorphoBlue/libraries/periphery/MorphoLib.sol";
+import { console } from "@forge-std/Test.sol";
```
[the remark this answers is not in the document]
0xEinCodes: This has been removed.

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueHelperLogic.sol
0xEinCodes: Interesting, I'll leave it out for now but it is noted.

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueCollateralAdaptor.sol
0xEinCodes: Thank for you pointing this out, we are looking into it :)

0xEinCodes:
> Do you have the full trace about this issue?
> 
> https://github.com/PeggyJV/cellar-contracts/blob/0b7a5f6074504b3563868c79c6a424cdaa691a7d/test/testAdaptors/MorphoBlue/MorphoBlueSupplyAdaptor.t.sol#L288

To replicate this error, simply comment out [this line](https://github.com/PeggyJV/cellar-contracts/blob/b639f56354e6a69f6dc644e6196ec3eb0c791b7a/test/testAdaptors/MorphoBlue/MorphoBlueSupplyAdaptor.t.sol#L288) and increase verbosity on the forge test. From digging into it a bit more, it seems that it is failing within the accounting check that occurs with every [`cellar.withdraw()` where it checks the balance of each `position` within a cellar via 
`getTotalAssetsAndTotalSupply()`](https://github.com/PeggyJV/cellar-contracts/blob/b639f56354e6a69f6dc644e6196ec3eb0c791b7a/src/base/Cellar.sol#L830) . So it ultimately queries `balanceOf()` within each cellar position, which carries out the `balanceOf()` implementation logic of each respective adaptor. So I think it is failing possibly because I am carrying out lossy accounting in my `balanceOf()` calculation for the `MorphoBlueSupplyAdaptor` seen [here](https://github.com/PeggyJV/cellar-contracts/blob/b639f56354e6a69f6dc644e6196ec3eb0c791b7a/src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueHelperLogic.sol#L61).

image

0xEinCodes:
Thank you for your reviews Morpho Team! I have carried out most change requests, but am still working on the following:

1. [Replace adaptorData with MarketParams instead of Id to save on gas](https://github.com/PeggyJV/cellar-contracts/pull/154#discussion_r1445751245)
2. [Further factorize MorphoBlueCollateralAdaptor addCollateral() and deposit() - this was factorized a bit already but it can be factorized more if we implement MarketParams as adaptorData](https://github.com/PeggyJV/cellar-contracts/pull/154#discussion_r1447928869)
3. [Resolving possible lossyness causing reverts in withdraw accounting checks with the supply adaptor seen here](https://github.com/PeggyJV/cellar-contracts/pull/154#issuecomment-1885999903)

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueSupplyAdaptor.sol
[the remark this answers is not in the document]
Rubilmax: Indeed, but:
1. it doesn't guarantee no dust shares or assets is left on Blue, while withdraw all the shares does
2. it requires you to accrue interest upfront to calculate `assets`, which costs a little (insignificant) gas

So @Jean-Grimal 's suggestion would be:

```suggestion
            morphoBlue.withdraw(market, 0, morphoBlue.supplyShares(_id, address(this)), address(this), address(this));
```

Jean-Grimal:
Hi @0xEinCodes, I investigated [this issue](https://github.com/PeggyJV/cellar-contracts/pull/154#issuecomment-1885999903), and found that the code reverts here, in the `Cellar` contract :


Capture d’écran 2024-01-11 à 16 55 29

I run some tests and got these logs :

Capture d’écran 2024-01-11 à 16 48 30

The owner of the shares minted by the initial deposit is the sender of the deployment tx of the cellar, and so in this case it is the `deployer` contract (address `0xFE99b7Adbe4D9344fD1f06e973E8c8D275E968a` in the logs).

However, it is the test contract (address `0x7FA9385bE102ac3EAc297483Dd6233D62b3e1496` in the logs) that deposits then withdraws from the cellar on behalf of itself. This address never was the owner of the initial shares, and only mints `assets` shares with the deposit (there is a 1:1 conversion rate between assets and shares), therefore can't redeem more than `asset` shares (i.e can't withdraw more than `assets` assets).

So that's why here the withdrawal of `(assets / 2) + 2` (knowing that `(assets / 2)` has already been withdrawn), fails.

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueSupplyAdaptor.sol
```diff
@@ -0,0 +1,234 @@
...
+     * @param adaptorData adaptor data containing the abi encoded Morpho Blue market.
+     * @return withdrawableSupply liquid amount of `loanToken` cellar has lent to specified MB market.
+     */
+    function withdrawableFrom(
+        bytes memory adaptorData,
+        bytes memory
+    ) public view override returns (uint256 withdrawableSupply) {
+        MarketParams memory market = abi.decode(adaptorData, (MarketParams));
+        (uint256 totalSupplyAssets, , uint256 totalBorrowAssets, ) = morphoBlue.expectedMarketBalances(market);
+        if (totalBorrowAssets == totalSupplyAssets) return 0;
```
crispymangoes, line 110: Is it ever possible for totalBorrowAssets to be greater than totalSupplyAssets? If so this should probably be `if (totalBorrowAssets >= totalSupplyAssets)`

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueSupplyAdaptor.sol
```diff
@@ -0,0 +1,234 @@
...
+    /**
+     * @notice Returns the amount of loanToken that can be withdrawn.
+     * @dev Compares loanToken supplied to loanToken borrowed to check for liquidity.
+     *      - If loanToken balance is greater than liquidity available, it returns the amount available.
+     * @param adaptorData adaptor data containing the abi encoded Morpho Blue market.
+     * @return withdrawableSupply liquid amount of `loanToken` cellar has lent to specified MB market.
+     */
+    function withdrawableFrom(
+        bytes memory adaptorData,
+        bytes memory
```
crispymangoes, line 106: Can you add `isLiquid` logic to the configuration data?

crispymangoes (changes requested):
Just a couple small changes, then I think we just run through the code on a call, and we are gucci :)

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueSupplyAdaptor.sol
0xEinCodes: The Morpho team actually pointed out to make this `==` vs `>=` since `totalBorrowAssets` cannot be greater than `totalSupplyAssets` as per [here](https://github.com/PeggyJV/cellar-contracts/pull/154#discussion_r1445835934).

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueSupplyAdaptor.sol
0xEinCodes: Done!

0xEinCodes:
All change requests have been accommodated/resolved. Thank you all for your time conducting PR Reviews! @crispymangoes and I are going to do a walk-through tomorrow before handing it off to the auditors. We are excited to get this to audit and onward!

test/resources/AdaptorHelperFunctions.sol
```diff
@@ -287,6 +294,65 @@ contract AdaptorHelperFunctions {
...
+        uint256 _assets
+    ) internal pure returns (bytes memory) {
+        return abi.encodeWithSelector(MorphoBlueSupplyAdaptor.withdrawFromMorphoBlue.selector, _market, _assets);
+    }
+
+    function _createBytesDataToAccrueInterestToMorphoBlueSupplyAdaptor(
+        MarketParams memory _market
+    ) internal pure returns (bytes memory) {
+        return abi.encodeWithSelector(MorphoBlueSupplyAdaptor.accrueInterest.selector, _market);
+    }
```
Rubilmax, line 318: What's the difference between this one and `_createBytesDataToAccrueInterestToMorphoBlue`?

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueCollateralAdaptor.sol
```diff
@@ -0,0 +1,222 @@
...
+        ERC20 collateralToken = ERC20(market.collateralToken);
+        _addCollateral(market, assets, collateralToken);
+    }
+
+    /**
+     * @notice User withdraws are NOT allowed from this position.
+     * NOTE: collateral withdrawal calls directly from users disallowed for now.
+     */
+    function withdraw(uint256, address, bytes memory, bytes memory) public pure override {
+        revert BaseAdaptor__UserWithdrawsNotAllowed();
```
Rubilmax, line 94: The user can deposit using `deposit` or `addCollateral`, but can only withdraw using `removeCollateral`. Is it intended? Why?

Jean-Grimal (approved):

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueDebtAdaptor.sol
```diff
@@ -130,41 +127,44 @@ contract MorphoBlueDebtAdaptor is BaseAdaptor, MorphoBlueHelperLogic {
...
-        uint256 sharesToRepay = debtAmountToRepay.toSharesUp(totalBorrowAssets, totalBorrowShares);
-        uint256 sharesAccToMorphoBlue = morphoBlue.borrowShares(_id, address(this));
-        if (sharesAccToMorphoBlue < sharesToRepay) {
-            sharesToRepay = sharesAccToMorphoBlue;
-        }
         tokenToRepay.safeApprove(address(morphoBlue), debtAmountToRepay);
-        _repayAsset(market, sharesToRepay, address(this));
+
+        uint256 totalBorrowAssets = morphoBlue.totalBorrowAssets(id);
+        uint256 totalBorrowShares = morphoBlue.totalBorrowShares(id);
```
MerlinEgalite, line 158: same suggestion here you could use a more efficient helper

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueCollateralAdaptor.sol
crispymangoes: So user deposits will go through the adaptors `deposit` function. ie When a user deposits into the vault, the money they deposited can be directly moved to morpho blue in the same TX. 

Where as `addCollateral` is how a strategist rebalancing the vault would add more collateral to morpho blue. 

We do not support user withdraws from morpho blue collateral bc withdraws will lower the health factor of the cellar, so we would need to calculate the health factor and make sure it is not below some minimum value. We could do it, but in practice we have found it much more gas efficient for users to just wait for the strategist to rebalance collateral out of the lending protocol first.

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueSupplyAdaptor.sol
crispymangoes: I think this is much safer, yes it is extremely unlikely for their to be a bug that would cause `totalBorrowAssets` to be greater than `totalSupplyAssets`, but if it ever does happen, leaving it as `==` means user withdraws will revert if we try to pull from the morpho blue supply position. The strategist would be able to recover the cellar from this state, but considering that making it `>=` would only add like 3 gas to the call, I think it is much better to go with the more robust design.

test/resources/AdaptorHelperFunctions.sol
crispymangoes: We have accrue interest in all the morpho blue adaptors just to make it easier for strategists, so they can call the function if need be without needing to switch a different adaptor, then go back to the adaptor they want to use.

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueSupplyAdaptor.sol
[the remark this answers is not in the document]
MerlinEgalite: aligned with the decision

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueHelperLogic.sol
```diff
@@ -93,12 +78,7 @@ contract MorphoBlueHelperLogic {
      */
     function _userBorrowBalance(Id _id, address _user) internal view returns (uint256) {
         Market memory market = morphoBlue.market(_id);
```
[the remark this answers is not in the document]
0xEinCodes: Good point, although we are comfortable paying the extra gas in this case :).

src/modules/adaptors/Morpho/MorphoBlue/MorphoBlueDebtAdaptor.sol
0xEinCodes: Another good point, due to the timeline and the smaller gas savings considered here, we are comfortable leaving this.

crispymangoes (approved):
Just rebase dev/macro-audit-16, and add missing natspec for new functions and its good to me :)

0xEinCodes:
I've finished making adjustments as per our conversation today @crispymangoes that went over the main changes made as per this PR Review. Crispy has rebased it to `dev/macro-audit-16`. I will squash and merge this in preparation for our audit kick-off today for Morpho Blue Adaptor scope :)

PR #164, opened 2023-12-11
Feat/illiquid erc20

crispymangoes:
Small edit to the ERC20Adaptor that allows for strategist to set erc20 positions as illiquid, or liquid, when adding them to the cellar.


0xEinCodes (approved):
Sweet, all looks good to me!

PR #166, opened 2023-12-12
Add initiator value to finishSolve and greatly simplify SimpleSolver …

crispymangoes:
…logic

### DIFF
--- a/test/WithdrawQueue.t.sol
+++ b/test/WithdrawQueue.t.sol
@@ -448,12 +448,20 @@ contract WithdrawQueueTest is MainnetStarterTest, AdaptorHelperFunctions, ISolve
         uint256[] memory expectedSharesToSolve,
         uint256[] memory expectedRequiredAssets
     ) internal {
-        (WithdrawQueue.SolveMetaData[] memory metaData, , ) = queue.viewSolveMetaData(share, users);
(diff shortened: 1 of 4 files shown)


test/WithdrawQueue.t.sol
```diff
@@ -739,20 +748,77 @@ contract WithdrawQueueTest is MainnetStarterTest, AdaptorHelperFunctions, ISolve
...
+        users[0] = attacker;
+
+        vm.startPrank(attacker);
+
+        // Call p2pSolve on reentrancy.
+        simpleSolverFunctionToReenter = 1;
+        vm.expectRevert(bytes("REENTRANCY"));
+        simpleSolver.redeemSolve(maliciousERC4626, users, 0, type(uint256).max);
+
+        // Call redeemSolve on reentrancy.
```
0xEinCodes, line 785: Should the called function be `p2pSolve` and `finishSolve` on the other two `expectRevert` tests here?

crispymangoes: No, so the way I am checking for reentrancy is the test contract has implemented a mock cellar interface, where the redeem function tries to reenter into the SimpleSolver(L817 in test file). But the only way to actually reenter is if the SimpleSolver calls the redeem function the test, which only happens if `redeemSolve` is used.

0xEinCodes: We discussed this further. Thanks!

PR #171, opened 2023-12-21
Feat/price router with sequencer

crispymangoes:


src/modules/price-router/PriceRouter.sol
```diff
@@ -533,6 +534,11 @@ contract PriceRouter is Ownable {
         return price;
     }
 
+    /**
+     * @notice If any safety checks needs to be run before pricing operations, they should be added here.
+     */
+    function _runPreFlightCheck() internal view virtual {}
```
0xEinCodes, line 540: Haha I like the name

src/modules/price-router/permutations/SequencerPriceRouter.sol
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * @notice Layer 2 chains that use sequencers, can have the sequencer go down. If this happens we do not want
+     *         to price assets, as the datafeeds could be stale, and need to be updated.
+     */
+    function _runPreFlightCheck() internal view override {
+        (
+            ,
+            /*uint80 roundID*/ int256 answer,
+            uint256 startedAt /*uint256 updatedAt*/ /*uint80 answeredInRound*/,
```
0xEinCodes, line 51: nit: Are you missing some commas here? It all compiles fine, so this is just semantic I think.

src/modules/price-router/permutations/SequencerPriceRouter.sol
```diff
@@ -0,0 +1,71 @@
...
+        ) = sequencerUptimeFeed.latestRoundData();
+
+        // This check should make TXs from L1 to L2 revert if someone tried interacting with the cellar while the sequencer is down.
+        // Answer == 0: Sequencer is up
+        // Answer == 1: Sequencer is down
+        if (answer == 1) {
+            revert SequencerPriceRouter__SequencerDown();
+        }
+
+        // TODO better understand why this adds a grace period, and why we use startedAt instead of updatedAt.
```
0xEinCodes, line 63: I am assuming this, but it sounds like `updatedAt` returns a nonzero value always where as `startedAt` returns a value that could be zero and is used to indicate whether or not the uptime status has changed or not. 

I'm just basing this off of [reading this part here,](https://docs.chain.link/data-feeds/l2-sequencer-feeds#:~:text=startedAt%3A%20This%20timestamp,the%20GRACE_PERIOD_TIME.) to be very sure one could dig into the `sequencerUptimeFeed` code.

test/testPriceRouter/SequencerPriceRouter.t.sol
```diff
@@ -0,0 +1,102 @@
...
+
+    function setUp() external {
+        // Setup forked environment.
+        string memory rpcKey = "ARBITRUM_RPC_URL";
+        uint256 blockNumber = 161730035;
+
+        uint256 forkId = vm.createFork(vm.envString(rpcKey), blockNumber);
+        vm.selectFork(forkId);
+
+        sequencerPriceRouter = new SequencerPriceRouter(address(this), 3_600, address(this), registry, WETH);
```
0xEinCodes, line 30: For the `_gracePeriod` constructor arg, what would be a fair value to use? This seems to be specific to the protocol and design situation being considered.

test/testPriceRouter/SequencerPriceRouter.t.sol
```diff
@@ -0,0 +1,102 @@
...
+        PriceRouter.ChainlinkDerivativeStorage memory stor;
+
+        PriceRouter.AssetSettings memory settings;
+
+        settings = PriceRouter.AssetSettings(1, USDC_USD_FEED);
+        sequencerPriceRouter.addAsset(USDC, settings, abi.encode(stor), 1e8);
+    }
+
+    function testSequencerUptimeFeedLogic() external {
+        // Sequencer is currently up so precising succeeds.
```
0xEinCodes, line 41: nit: spelling - I think you meant "pricing" not "precising"

test/testPriceRouter/SequencerPriceRouter.t.sol
```diff
@@ -0,0 +1,102 @@
...
+        mockStartedAt = block.timestamp - 3_601;
+
+        // Pricing calls now succeed.
+        sequencerPriceRouter.getPriceInUSD(USDC);
+    }
+
+    function testSequencerFeed() external {
+        // Setup forked environment.
+        string memory rpcKey = "ARBITRUM_RPC_URL";
+        uint256 blockNumber = 160355845;
```
0xEinCodes, line 72: May you please elaborate on this test? Is this simulating what actually was happening when the Arbitrum sequencer was down 6+ days ago?

Also, are you leaving the console.logs here for a reason?

At first I thought this test was to check how a fork of the Arbitrum network was actually performing during a time when the sequencer was down. If so, then setting this test address as the `sequencerPriceRouter` again and trying to use it to ensure that it performs the way we want it to (revert), is a good idea.

test/testPriceRouter/SequencerPriceRouter.t.sol
```diff
@@ -0,0 +1,102 @@
...
+
+        // PriceRouter.ChainlinkDerivativeStorage memory stor;
+
+        // PriceRouter.AssetSettings memory settings;
+
+        // settings = PriceRouter.AssetSettings(1, USDC_USD_FEED);
+        // sequencerPriceRouter.addAsset(USDC, settings, abi.encode(stor), 1e8);
+    }
+
+    int256 mockAnswer = type(int256).max;
```
0xEinCodes, line 89: nit: move these vars to the top of the test contract (just formatting)

test/testPriceRouter/SequencerPriceRouter.t.sol
```diff
@@ -0,0 +1,102 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.21;
+
+import { SequencerPriceRouter, PriceRouter, Registry, ERC20, IChainlinkAggregator } from "src/modules/price-router/permutations/SequencerPriceRouter.sol";
+import { Math } from "src/utils/Math.sol";
+
+import { Test, stdStorage, StdStorage, stdError, console } from "@forge-std/Test.sol";
+
+// TODO refactor to a multichain test once deploy is merged into branch.
```
0xEinCodes, line 9: What do you think about creating an open issue that has a list of reformatting and other TODOs for multichain tests, etc.? That way we can remove this TODO for now.

0xEinCodes (changes requested):
Looking great @crispymangoes! There are some change requests, more-so small things. I have a general question on your second test that you have.

Also, the tests cover ensuring that the `PriceRouter.sol` will revert, and thus other calls reliant on it from Cellars that users can call will revert too. This means that no txs from Sommelier Cellars to an L2 that has a sequencer down will be queued I guess? Queued in the Arbitrum or other L2 queue I mean, not the Sommelier `withdrawQueue`. 

If the above is true, is there a scenario / setup where strategists need to increase their health factor (via more collateral or repaying debt) on some cellar positions in those sequencer-down L2s?

src/modules/price-router/PriceRouter.sol
crispymangoes: Me too :)

src/modules/price-router/permutations/SequencerPriceRouter.sol
crispymangoes: Lol I tried fixing it but the linter made it look way worse so I just removed the inline comments

src/modules/price-router/permutations/SequencerPriceRouter.sol
crispymangoes: Yeah I asked chainlink for clarification, but based off that comment it sounds like it was implemented right. But we will wait to hear back from chainlink.

test/testPriceRouter/SequencerPriceRouter.t.sol
crispymangoes: Fixed!

test/testPriceRouter/SequencerPriceRouter.t.sol
crispymangoes: This test is still a work in progress. I am pretty sure that during this fork period the sequencer is down, but the console logged answer still show its up. Part of this is confirming with chainlink that I am using the correct sequencer uptime feed.

test/testPriceRouter/SequencerPriceRouter.t.sol
crispymangoes: Done

src/modules/price-router/permutations/SequencerPriceRouter.sol
0xEinCodes: That makes sense, we'll resolve it then :)

test/testPriceRouter/SequencerPriceRouter.t.sol
0xEinCodes: Sounds good. Resolving this here as you have this in progress already and are working on it but it will likely be after the Chainlink team returns back to us with feedback in the New Year.

test/testPriceRouter/SequencerPriceRouter.t.sol
0xEinCodes: sg2m

test/testPriceRouter/SequencerPriceRouter.t.sol
0xEinCodes: That makes sense. We also discussed the main question I had within the PR review comment about having the gracePeriod in lock-step with external protocols that L2 cellars interact with. We decided that some stagger would be okay for protocol integrations. We can touch on this more with Chainlink as they get back to us in the new year too.

0xEinCodes: I'll leave this unresolved so we can make sure to discuss it with Macro cause I'm curious about their thoughts on it too.

0xEinCodes (approved):

PR #172, opened 2024-01-03
Feat/multi asset deposit

crispymangoes:


test/CellarWithMultiAssetDeposit.t.sol
```diff
@@ -0,0 +1,331 @@
...
+        vm.label(strategist, "strategist");
+        // Approve cellar to spend all assets.
+        USDC.approve(address(cellar), type(uint256).max);
+        initialAssets = cellar.totalAssets();
+        initialShares = cellar.totalSupply();
+    }
+
+    // ========================================= HAPPY PATH TEST =========================================
+
+    // Can we accept the Curve LP tokens? I dont see why not.
```
0xEinCodes, line 114: Is this note needed still? I think we move it to an issue if we are considering adding in Curve LP tokens and testing with them. Unless we think we don't need to test with them in an extra way.

test/CellarWithMultiAssetDeposit.t.sol
```diff
@@ -0,0 +1,331 @@
...
+        USDT.safeApprove(address(cellar), assets);
+
+        bytes memory depositCallData = abi.encodeWithSelector(Cellar.deposit.selector, assets, address(this), USDT);
+
+        address(cellar).functionCall(depositCallData);
+
+        // Since share price is 1:1, and USDT is hardcoded to equal the same as USDC, below checks should pass.
+        assertEq(cellar.previewRedeem(1e6), 1e6, "Cellar share price should be 1.");
+
+        cellar.dropAlternativeAssetData(USDT);
```
0xEinCodes, line 152: Should we remove this and the commented out LoC below? You test dropping an asset later on in a different test I believe.

test/CellarWithMultiAssetDeposit.t.sol
```diff
@@ -0,0 +1,331 @@
...
+            "Cellar totalAssets should equal initial + new deposit."
+        );
+        assertEq(
+            cellar.totalSupply(),
+            initialAssets + assets,
+            "Cellar totalSupply should equal initial + new deposit."
+        ); // Because share price is 1:1.
+    }
+
+    function testDepositWithAlternativeAsset(uint256 assets) external {
```
0xEinCodes, line 136: Do you think it is worth it to test with an asset that is not 1:1? What about native ETH?

0xEinCodes (commented):
In general, this looks good. That said, I made one comment simply on adding tests for non-correlated assets. As well, can you remind me on:

- Why is `onlyOwner()` now a helper function instead of a modifier?
- Why is the `automation` and `viewPositionBalances` sections of the `Cellar.sol` removed?

test/CellarWithMultiAssetDeposit.t.sol
crispymangoes: onlyOwner was made an internal function to save on contract size. When you use modifiers the modifier code is copied to all functions it decorates. By using an internal function modifier code is only copied once.

automation and viewPositionBalances were removed because they were not being used, and we needed more contract space. Thinking about it, being able to deposit multiple assets is much more valuable than automation actions.

test/CellarWithMultiAssetDeposit.t.sol
crispymangoes: Yeah actually I am not sure why I drop it there, also why I kept that commented out LoC. I'll remove them.

test/CellarWithMultiAssetDeposit.t.sol
crispymangoes: Actually the 1:1 hard coding was removed, so we are already testing with assets that are not 1:1.

crispymangoes: We still dont support native ETH. Adding logic to support it, takes up too much contract space.

0xEinCodes: Sg2m

0xEinCodes (approved):
Approved after checking your responses. Nice job!

crispymangoes:
This PR is being closed as the code was merged into dev/macro-audit-16 which will be merged into main once the audit is complete.

PR #174, opened 2024-01-09
Add audit #14 report, where we skipped the #13

0xEinCodes:
## Core changes:

This PR encompasses adding in the latest smart contract audit (Audit # 14) carried out by Macro. This is technically audit # 13 but we skipped it in file numbering intentionally. See reports for the smart contracts covered.


crispymangoes (approved):
Looks good to me

PR #175, opened 2024-01-11
Fix/macro audit 15

crispymangoes:
Fix findings found in Macro Audit 15.

### DIFF
--- a/src/base/ERC4626SharePriceOracle.sol
+++ b/src/base/ERC4626SharePriceOracle.sol
@@ -138,7 +138,14 @@ contract ERC4626SharePriceOracle is AutomationCompatibleInterface {
      */
     event KillSwitchActivated(uint256 reportedAnswer, uint256 minAnswer, uint256 maxAnswer);
 
+    /**
+     * @notice Emitted when the upkeep is registered.
+     */
     event UpkeepRegistered(uint256 upkeepId, address forwarder);
+
+    /**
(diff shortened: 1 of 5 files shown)


0xEinCodes (approved):
Looks good to me, nice work @crispymangoes. The only comment I have is about the high-level design:
- Instead of going with the revised `ERC4626SharePriceOracle.sol`, we could go with a separate version altogether for a `ERC4626SharePriceOracle.sol` that checks if a `sequencerUptimeFeed` is used or not. We could have `internal virtual` setters / getters for the uptime feed checks and set up the oracles that way so they are consistently operating with the same functionality / design with extensions though instead of changing the single file altogether. Either way works for me, but it's just food for thought.

Oh, do we want to test with the actual uptimeFeed for Arbitrum and other related L2s within our forked test suite? I understand that these will be trialled with test cellars too on respective prod networks before full production use.

crispymangoes:
So the reason why I didnt want to go with another version of the share price oracle contract, is so that we dont need multiple versions of cross chain share price oracles, we will just need 1. Also I think it is likely that mainnet oracles relaying an L2 cellars share price will use this logic in them to check the state of the L2 sequencer.

As for adding a real test, I think having a mock one is fine since mimicing the state of the real data feed is so easy. Plus the last arbiturm outage was not properly recorded by the sequencer uptime feed.

PR #179, opened 2024-01-16
Feat/separate deposit and multi asset deposit

crispymangoes:


0xEinCodes (approved):
Overall looks good to me, nice work on this! These changes were a result of discussion with the auditors iirc. My review is likely not needed or at least can wait until they add further review since they've already started.

PR #188, opened 2024-02-21
Fix eETHExtension

0xEinCodes:
As per conversations with Crispy, there was a bug in eETH extension. This has been caught and reviewed with external auditors, and pricing using this extension code has been revised before this fix. This fix reflects those changes.

Deployment that was corrected before this PR (so it is good): https://etherscan.io/address/0xE5Dc10b6474D6Ac66388d946F2E6B91E4759B8AC#code 

The problem was that `getPriceInUSD` was multiplying the weETH price by the eETH/weETH rate, when it really needs to be dividing out the rate.

## Summary by CodeRabbit

- **New Features**
	- Updated the logic for calculating `eETH` prices in USD to ensure more accurate pricing for `eETH` assets.


crispymangoes (approved):
Looks correct to me!

